PR #1, opened 2015-10-08
Added README

sandlerben:
r @kyle-rosenbluth

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,44 @@
+# Clean Air Council
+
+## Running the app
+
+1. Clone the repo
+
+    ```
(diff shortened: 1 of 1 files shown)


kyle-rosenbluth:
Looks good. What do you think of breaking it up into 2 sections for first install and subsequent use? I.e. I don't need to clone the repo every time I run it, but I would want to activate the virtualenv and run the server every time.

sandlerben:
How does it look now @aharelick @kyle-rosenbluth

kyle-rosenbluth:
:+1:

PR #2, opened 2015-10-11
Created the basic maps.html

sanjayss34:
I've created a new html file for the basic map functionality and the functionality with entering an address in a search box and re-centering the map to the location of that address. The index.html file is slightly edited, but it should really contain the same code as the initial file (edits were mainly/only whitespace related).


sandlerben:
Thanks @sanjayss34. Will review tonight!

app/templates/main/index.html
```diff
@@ -14,4 +14,5 @@
 
         <p>{{ lipsum(6) }}</p>
     </div>
-{% endblock %}
\ No newline at end of file
+
```
sandlerben: Nit: Can you revert this file?

sandlerben:
A few general comments:
1. It looks like there isn't any indentation in your map.html file. Indentation is really important for readable html and javascript! :bowtie: 
2. Can you explain why you and @apiatski decided to use the Google Maps API? I am sure it's the right decision, but I am curious.
3. Before we merge this in, let's hook it up to an actual flask route in main/views.py and make maps.html extend base.html. Doing these two things will allow us to take advantage of all the fancy styling, asset management, etc. Moreover, any changes we make in base.html will be reflected here. You can read more about "child templates" [here](http://jinja.pocoo.org/docs/dev/templates/#child-template). If you have trouble figuring out how to do this, either @apiatski or I can help.

app/templates/main/map.html
```diff
@@ -0,0 +1,72 @@
+<html>
+<head>
+<script src="http://maps.googleapis.com/maps/api/js"></script>
```
sandlerben: If we can download the full API, we should move this logic to app/assets.py. If we want to always pull the most updated API code, we can add this to app/templates/partials/_head.html

app/templates/main/map.html
```diff
@@ -0,0 +1,72 @@
...
+<script>
+geocoder = new google.maps.Geocoder();
+address = document.getElementById("address");
```
sandlerben: We should try to use jQuery for getting text from a text box. In general, jQuery abstracts away a lot of cross-browser incompatibility weirdness. While this code will almost certainly work, I think using jQuery is a better practice.

app/templates/main/map.html
```diff
@@ -0,0 +1,72 @@
...
+function initialize() {
+var initial_coords = new google.maps.LatLng(39.952, -75.195);
```
sandlerben: Why are we hard-coding these values?

app/templates/main/map.html
```diff
@@ -0,0 +1,72 @@
...
+var mapProp = {
+center: initial_coords,
+zoom:17,
```
sandlerben: What does hard-coding a zoom value of 17 mean?

app/templates/main/map.html
```diff
@@ -0,0 +1,72 @@
...
+mapTypeId: google.maps.MapTypeId.ROADMAP
+};
+map=new google.maps.Map(document.getElementById("googleMap"),mapProp);
+}
+function reinitialize() {
+var initial_coords = new google.maps.LatLng(39.952, -75.195);
```
sandlerben: Same as above

app/templates/main/map.html
```diff
@@ -0,0 +1,72 @@
...
+var mapProp = {
+center: initial_coords,
+zoom:17,
```
sandlerben: Same as above

app/templates/main/map.html
```diff
@@ -0,0 +1,72 @@
...
+var map=new google.maps.Map(document.getElementById("googleMap"),mapProp);
+if (geocoder) {
+console.log(address);
+geocoder.geocode( { 'address': address}, function(results, status) {
+if (status == google.maps.GeocoderStatus.OK) {
+if (status != google.maps.GeocoderStatus.ZERO_RESULTS) {
+map.setCenter(results[0].geometry.location);
+}
+else {
+alert("Address not found!");
```
sandlerben: Let's try to use flashing and not alerts. You can see the html you need to output to trigger a flash in app/templates/partials/_flashes.html. It would be great if you could define a JavaScript function which outputs that exact html (since Max already styled it). To see an example of what I mean, check out https://github.com/hack4impact/2015-project-tester/blob/master/app/templates/submission.html#L46

To see this in practice, go to http://project-testing.herokuapp.com/ and give it a URL. Notice how error messages are "flashed" to the top of the webpage. When you first load the page, you see the message "Messages will show up here. Built by Ben Sandler."

app/templates/main/map.html
```diff
@@ -0,0 +1,72 @@
...
+}
+}
+else {
+alert("Geocode not successful - " + status);
+}
+});
+}
+}
+
+google.maps.event.addDomListener(window, 'load', initialize);
```
sandlerben: Once again, let's use JQuery for this. Look for an onClick listener. Defining a form with JavaScript actions is less elegant I think.

app/templates/main/map.html
```diff
@@ -0,0 +1,72 @@
...
+</script>
+<form name = "addressForm" action="JavaScript:update_center()">
+    <input type="text" name="address" id="address" value="Enter Address">
+    <br><input type="submit" value="Submit"><br>
+</form>
+</head>
+<body>
+<div id="googleMap" style="width:500px;height:380px;"></div>
```
sandlerben: We shouldn't hard-code the size of the map. To be responsive, it should fill the webpage. Also, let's try to move style like this to CSS.

sandlerben:
Overall, this is a solid first start. You also win the medal for first pull request :trophy: 

@nancywong do you mind reviewing this, too? You have experience with front-end and might have some comments.

app/templates/main/map.html
```diff
@@ -0,0 +1,72 @@
+<html>
+<head>
+<script src="http://maps.googleapis.com/maps/api/js"></script>
+<script>
```
rileynwong: I would put this in a separate .js file and folder for your static javascript, like in `app/templates/js` or `app/static/js`. Then include the script here with `<script src ="path/to/file.js"></script>`

app/templates/main/map.html
rileynwong: +1 to getting rid of "magic numbers" for zoom and lat/long (see http://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad)

app/templates/main/map.html
```diff
@@ -0,0 +1,72 @@
...
+geocoder = new google.maps.Geocoder();
```
rileynwong: Could you move the js to a separate file and add indentation and spacing?

app/templates/main/map.html
rileynwong: anything that uses the `style` tag should be stored in a scss file (since we are using SASS as our css framework) - maybe create a `map.scss` file and put it in `app/assets/styles` containing your styling

app/templates/main/map.html
```diff
@@ -0,0 +1,72 @@
...
+if (status == google.maps.GeocoderStatus.OK) {
+if (status != google.maps.GeocoderStatus.ZERO_RESULTS) {
+map.setCenter(results[0].geometry.location);
+}
+else {
+alert("Address not found!");
+}
+}
+else {
+alert("Geocode not successful - " + status);
```
rileynwong: Instead of nesting these if-statements, could you do them sequentially? Or at least modify the logic s.t. the flash/alert logic directly follows the conditional.
e.g.

```
if !(geocoderStatus) {
    alert("Geocode not successful")
}

if!(zeroResults) {
    alert("No address")
}
```

if they must be nested, then put the error alert _before_ entering a new conditional

app/templates/main/map.html
```diff
@@ -0,0 +1,72 @@
...
+google.maps.event.addDomListener(window, 'load', initialize);
+</script>
+<form name = "addressForm" action="JavaScript:update_center()">
+    <input type="text" name="address" id="address" value="Enter Address">
+    <br><input type="submit" value="Submit"><br>
+</form>
+</head>
+<body>
+<div id="googleMap" style="width:500px;height:380px;"></div>
+</body>
```
rileynwong: I might add indentations and spacing to the HTML tags for visual clarity

sandlerben:
or*

sandlerben:
Also, by "you" I mean Alex

sandlerben:
@sanjayss34 @apiatski when you guys are ready for another round of reviews let @nancywong and me know.

README.md
```diff
@@ -21,7 +21,7 @@
 
     ```
     $ pip install -r requirements/common.txt
-    $ pip install -r requirements/dev.txt
+    $ pip install -r requirements/dev.  txt
```
sandlerben: Why did this change?

sanjayss34: I'm not really sure. Do you know which one is the correct version? (sorry for the late response)

app/static/js/initialize_map.js
```diff
@@ -0,0 +1,155 @@
...
+INITIAL_CENTER_LONG = -75.195;
+initial_coords = new google.maps.LatLng(INITIAL_CENTER_LAT, INITIAL_CENTER_LONG);
+//Initial map zoom value
+ZOOM = 17;
+//An integer for the offset between the actual window size and the
+//JavaScript-determined Window size in order to size the map
+WINDOW_OFFSET = 70
+
+//Bounds for the Date Slider. End date is current date.
+var BOUNDS_MIN = new Date(2015, 0, 1);
```
kongsally: The min date should depend on the earliest date from the incident reports, and if null, to (2015, 1, 1)

app/templates/main/map.html
```diff
@@ -0,0 +1,61 @@
...
+    <body>
+     <div class ="column">
+         <div id="wrapper">
+             <div class="row">
+                <div id="googleMap"></div>
+                           <div class="ui form">
+              <div class="fields">
+                <div class="three wide field">
+                  <label>License Plate Number</label>
+                  <input type="text" placeholder="XXX-XXX">
```
kongsally: Maybe change it to eg.MG-1234E
Might be misleading since the truck license numbers like the ones in the sample have 7 characters

app/templates/main/map.html
```diff
@@ -0,0 +1,61 @@
...
+                </div>
+                <div class="three wide field">
+                  <label>Vehicle ID</label>
+                  <input type="text" placeholder="enter vehicle ID">
```
kongsally: I think it would be good to keep the placeholder consistent. So maybe:
"eg. 105014"

app/templates/main/map.html
```diff
@@ -0,0 +1,61 @@
...
+                </div>
+              </div>
+              <div class="fields">
+                <div class="six wide field">
+                    <label>Notes </label>
+                  <input type="text" placeholder="(e.g. 'Was the driver sleeping?')">
```
kongsally: e.g. Driver sleeping

app/templates/main/map.html
```diff
@@ -0,0 +1,61 @@
...
+                </div>
+                <div class="three wide field">
+                    <label>Idling Duration </label>
```
kongsally: label should have (in minutes) and the input should be just a number,
so maybe eg. 10

app/static/styles/map.css
```diff
@@ -0,0 +1,35 @@
+
```
kongsally: change it to map.scss so that we can use some common sass variables to style it. You can read up on it here (http://sass-lang.com/guide)

app/templates/main/map.html
```diff
@@ -0,0 +1,61 @@
...
+                  <input type="number" placeholder="~X minutes">
+                </div>
+            </div>
+
+
+             </div>
+    	    <div id="slider"></div>
+            <form name = "addressForm" id = "addressForm">
```
kongsally: wrap it around a semantic ui search http://semantic-ui.com/modules/search.html or try adding the class to the form and see if that changes it too

README.md
sandlerben: No problem. I would just reset this file. You can do that with git ([read here](https://github.com/blog/2019-how-to-undo-almost-anything-with-git)) or (much easier) just delete the space in the file and recommit

app/main/views.py
```diff
@@ -1,7 +1,39 @@
...
+    license_plates = []
+    latitudes = []
+    longitudes = []
+    dates = []
+    durations = []
+    agencies = []
+    pictures = []
+    descriptions = []
+
+    for ir in models.IncidentReport.query.all():
```
sandlerben: We should be more specific about what we want to import from models.

``` python
from models import IncidentReport

for ir in IncidentReport.query.all():
```

app/main/views.py
```diff
@@ -1,7 +1,39 @@
...
+from app import models, db
 
 @main.route('/')
 def index():
     return render_template('main/index.html')
+
+
+@main.route('/map')
+def get_map():
+    #Get information on all Incident Reports in the db, and pass to
```
sandlerben: You should use the more traditional docstring format. Check out most other functions with comments and/or [read this](https://www.python.org/dev/peps/pep-0257/)

app/main/views.py
```diff
@@ -1,7 +1,39 @@
...
+        vehicle_ids.append(str(ir.vehicle_id))
+        license_plates.append(str(ir.license_plate))
+
+#Pass latitudes, longitudes as floats so they can be used to fix
+        #locations of map markers
+        latitudes.append(float((ir.location).latitude))
+        longitudes.append(float((ir.location).longitude))
+        dates.append(str(ir.date))
+        durations.append(str(ir.duration))
+        print('vehicle id:', ir.vehicle_id)
```
sandlerben: Was this for debugging?

app/static/js/initialize_map.js
```diff
@@ -0,0 +1,155 @@
+geocoder = new google.maps.Geocoder();
+address = document.getElementById("address");
+//Initial map center coordinates
+INITIAL_CENTER_LAT = 39.952;
```
sandlerben: Is hardcoding center coordinates necessary if we are just going to auto-recenter the map?

app/static/js/initialize_map.js
```diff
@@ -0,0 +1,155 @@
...
+INITIAL_CENTER_LONG = -75.195;
+initial_coords = new google.maps.LatLng(INITIAL_CENTER_LAT, INITIAL_CENTER_LONG);
+//Initial map zoom value
+ZOOM = 17;
```
sandlerben: Hardcoding ZOOM is temporary right? For anything here which is temporary, add a `// TODO:` comment

app/static/js/initialize_map.js
```diff
@@ -0,0 +1,155 @@
...
+//An integer for the offset between the actual window size and the
+//JavaScript-determined Window size in order to size the map
+WINDOW_OFFSET = 70
```
sandlerben: Why this value? Will this work reliably across devices?

app/main/views.py
```diff
@@ -1,7 +1,39 @@
...
+        agencies.append(str(ir.agency.name))
+        pictures.append(ir.picture_url)
+        descriptions.append(str(ir.description))
+
+    return render_template('main/map.html', vehicle_ids=vehicle_ids, license_plates=license_plates, latitudes=latitudes, longitudes=longitudes, dates=dates, durations=durations, agencies=agencies, pictures=pictures, descriptions=descriptions)
```
sandlerben: See comment below!

app/static/js/initialize_map.js
```diff
@@ -0,0 +1,155 @@
...
+            
+            //Information presented when marker is clicked
+            var contentString = '<div id="content">' +
+                '<div id="siteNotice">' +
+                '</div>' +
+                '<h1 id="firstHeading" class="firstHeading">Vehicle ID: ' + vehicle_ids[2*i+1] +'</h1>' +
+                '<div id="bodyContent">' +
+                '<p>License Plate: ' + license_plates[2*i+1] + '</p>' +
+                '<p>Date: ' + dates[2*i+1] + '</p>' +
+                '<p>Duration: ' + durations[2*i+1] + '</p>' +
```
sandlerben: This seems like a bit hacky/messy. I propose a more elegant way to do this

Just pass a list of `IncidentReport`s to the template

``` python
incident_reports = IncidentReport.query.all()
return render_template('main/map.html', incident_reports=incident_reports)
```

Then add the markers like this

``` javascript
{% for report in incident_reports %}
var marker = new google.maps.Marker({
               //Use latitude and longitude values from incident report
               //to set position of marker
               position:{lat: '{{ report.location.latitude }}', lng: '{{ report.location.longitude }}' }
});
marker.setMap(map);
var contentString = '<div id="content">' +
'<div id="siteNotice">' +
'</div>' +
'<h1 id="firstHeading" class="firstHeading">Vehicle ID: ' + '{{ report.vehicle_id }}'
'</h1>' +
...
{% endfor %}
```

Note: For this to work, we will at least have to generate the markers in the maps.html file. This is a perfectly fine tradeoff because it will make code nicer in my opinion.

sandlerben:
Still haven't gotten through the entire PR but looks pretty good.

If you need any help refactoring the marker generating code let me know!

app/static/js/initialize_map.js
```diff
@@ -0,0 +1,285 @@
+geocoder = new google.maps.Geocoder();
+address = document.getElementById("address");
+
+//Global array for map markers
```
sandlerben: Nit: Commetns will be easier to read if you have spaces between `//` and comment text like

``` javascript
// Global array for map markers
```

sandlerben:
Some of my comments are hidden because they are on "outdated diffs." Make sure you go on the "conversation" page to see them since a few are still applicable.

app/static/js/initialize_map.js
```diff
@@ -0,0 +1,285 @@
...
+    });
+    $("#slider").bind("valuesChanged", function(e, data) {
+        console.log("Values just changed. min: " + String(data.values.min).substring(4, 15) + " max: " + String(data.values.max).substring(4, 15));
+        var beginYear = parseInt(String(data.values.min).substring(11, 15), 10);
+        var beginDay = parseInt(String(data.values.min).substring(8, 10), 10);
+        var endYear = parseInt(String(data.values.max).substring(11, 15), 10);
+        var endDay = parseInt(String(data.values.min).substring(8, 10), 10);
+        var beginMonth = 0;
+        var endMonth = 0;
+        switch(String(data.values.min).substring(4, 7)) {
```
sandlerben: Consider doing it like this

``` javascript
monthObj = {
   "Jan": 0,
   "Fed": 1,
    ...
};

beginMonth = monthObj[String(data.values.min).substring(4, 7)];
endMonth = monthObj[String(data.values.max).substring(4, 7)];
```

app/static/js/initialize_map.js
```diff
@@ -0,0 +1,285 @@
...
+                beginMonth = 9;
+                break;
+            case "Nov":
+                beginMonth = 10;
+                break;
+            case "Dec":
+                beginMonth = 11;
+                break;
+        }
+        switch(String(data.values.max).substring(4, 7)) {
```
sandlerben: If you were to keep using a switch statement you should refactor this repeated code into its own function

app/static/js/initialize_map.js
sanjayss34: Yes, I think so because if we are denied access to user location (the browser asks for permission from the user to use location), then we still need to center the map somewhere. Do you have other suggestions?

app/static/js/initialize_map.js
sanjayss34: What should we do rather than hardcoding zoom? Note that the user still has the ability to adjust the zoom value from the ui.

app/static/js/initialize_map.js
sanjayss34: Just to clarify: so you're saying add a
<script> section in the html file, and put the marker-generating code there? If so, would the date slider code still go in the same place, as the date slider has to interact with the marker-adding function?

sandlerben:
In the future, you may want to squash similar commits (e.g. "pep8 stuff" 1-4) into one commit using `git commit --amend`

sandlerben:
Over slack, you mentioned that a few functions use `addMarkers`. Can you explain what they use it for?

app/templates/main/map.html
```diff
@@ -0,0 +1,53 @@
+{% extends 'layouts/base.html' %}
+{% block content %}
+<link rel="stylesheet" href="static/styles/iThing.css" type="text/css" />
+<link rel="stylesheet" href="static/styles/map.css" type="text/css" />
+<script src="static/js/jquery-ui.js"></script>
+<script src="static/js/jQDateRangeSlider-min.js"></script>
+<script src="static/js/initialize_map.js"></script> 
+<meta id="vehicle_ids" data-name="{{vehicle_ids}}">
```
sandlerben: Hmm, I've never seen it done like this. Usually, I've seen jinja -> javascript done with the jinja `json` (or something like this) functions.

This won't matter though assuming you make the changes I described earlier.

sandlerben:
Ok here is what I would do.
- Keep all the javascript in a separate file like it is now. 
- Update the `addMarkers` function to take a javascript object which contains all the data about reports. Instead of `addMarkers` reading report information from `$(<some field>).data` it will read that info from the reports javascript object.
- You will a slight bit of javascript in the maps.html template which reads the reports data from jinja and converts it to a native javascript object. Once you have this javascript object, call the `addMarkers` function with it.

Here is a sketch of how this might looks. Note: I will start the serialize function (and you can modify it to meet your needs).

**The view**

``` python
import json

@app.route('/maps')
def my_view():
    reports_data = [report.serialize() for report in IncidentReport.query.all()]
    return render_template('maps.html', reports_data=json.dumps(reports_data))
```

**The template**

``` javascript
<script type="text/javascript">
    addMarkers({{ reports_data|safe }}, <other arguments>);
</script>
```

sandlerben:
ACTUALLY ignore that comment.

Here is what you should really do.
- Define a new type of object called a `MarkerWrapper` in your javascript file

``` javascript
    function MarkerWrapper(actualMarker, incidentDate, contentString) {
        this.actualMarker = actualMarker; // actualMarker is of type google.maps.Marker, you should instantiate it outside of this constructor just like you do here https://github.com/hack4impact/clean-air-council/pull/2/files#diff-de047b44d935882c6e5ccd52a76f3ff4R63
        this.incidentDate = incidentDate; // a Date object, like here https://github.com/hack4impact/clean-air-council/pull/2/files#diff-de047b44d935882c6e5ccd52a76f3ff4R75
        this.contentString = contentString; // same as https://github.com/hack4impact/clean-air-council/pull/2/files#diff-de047b44d935882c6e5ccd52a76f3ff4R88
    }
```
- Add code like this in your html file to create a list of MarkerWrappers and the map

``` javascript
<script type="text/javascript">
    function initalizeMapAndMarkers() {
        // <generate map here just like before>



        markerWrappers = [];
        {% for report in incident_reports %}
        var marker = new google.maps.Marker({
                // Use latitude and longitude values from incident report
                // to set position of marker
                position:{'{{ report.latitude }}', '{{ report.longitude }}'}
        });
        incidentDate = // get date info from jinja
        marker.setMap(map);
        var contentString = '<div id="content">' +
                '<div id="siteNotice">' +
                '</div>' +
                '<h1 id="firstHeading" class="firstHeading">Vehicle ID: ' + '{{ report.vehicle_id }}' +'</h1>' + ...
        markerWrapper = new MarkerWrapper(marker, incidentDate, contentString);
        markerWrappers.push(markerWrapper);
        var infoWindow = new google.maps.InfoWindow({
                content: contentString
            });

            // Add click listener to marker for displaying infoWindow
            marker.addListener('click', function() {
                infoWindow.open(map, marker);
            });
        {% endfor %}
        storeMarkerState(markerWrappers, map);   
    }
</script>
```
- In your javascript files, define a function like this

``` javascript
var globalMarkerWrappers = null; // global variable
var globalMap = null; // global variable
function storeMarkerState(markerWrappers, map) {
    globalMarkerWrappers = markerWrappers;
    globalMap = map;
}
```
- Also in your javascript files, define the dateSlider functions. Now, they should iterate through the list of `MarkerWrapper` in globalMarkerWrappers. For each one, if `wrapper.incidentDate` is valid make sure it is ON the map and if it is not valid make sure it is OFF the map

app/__init__.py
```diff
@@ -9,7 +9,7 @@
 from flask.ext.wtf import CsrfProtect
 from flask.ext.compress import Compress
 from config import config
-from assets import app_css, app_js, vendor_css, vendor_js
+from assets import app_css, app_js, vendor_css, vendor_js, iThing_css, jqueryui_js, jQDateRangeSlidermin_js
```
sandlerben: I think you are doing this wrong. You can add those files themselves to `vendor_css` and `vendor_js` in assets.py

app/__init__.py
```diff
@@ -49,6 +49,9 @@ def create_app(config_name):
     assets_env.register('app_js', app_js)
     assets_env.register('vendor_css', vendor_css)
     assets_env.register('vendor_js', vendor_js)
+    assets_env.register('iThing_css', iThing_css)
```
sandlerben: See above

app/assets.py
```diff
@@ -24,3 +24,18 @@
     filters='jsmin',
     output='scripts/vendor.js'
 )
+
+iThing_css = Bundle(
+    'iThing.css',
```
sandlerben: Just move this stuff under `vendor_css` and `vendor_js`

app/static/js/initialize_map.js
```diff
@@ -0,0 +1,119 @@
+geocoder = new google.maps.Geocoder();
```
sandlerben: Can we rename this file to something like `map.js` since it includes more than initializing the map

app/static/js/initialize_map.js
```diff
@@ -0,0 +1,119 @@
...
+address = document.getElementById("address");
```
sandlerben: Use jquery

app/static/js/initialize_map.js
```diff
@@ -0,0 +1,119 @@
...
+
+// Global array for map markers
+var markers = []
```
sandlerben: What is this needed for?

app/static/js/initialize_map.js
```diff
@@ -0,0 +1,119 @@
...
+
+// Global Marker Wrappers and Map
+var globalMarkerWrappers = null;
+var globalMap = null;
+
+// Initial map center coordinates
+INITIAL_CENTER_LAT = 39.952;
```
sandlerben: You should recenter using something like [this](http://stackoverflow.com/questions/19304574/center-set-zoom-of-map-to-cover-all-markers-visible-markers)

That way you won't need to hardcode a center

app/static/js/initialize_map.js
```diff
@@ -0,0 +1,119 @@
...
+INITIAL_CENTER_LONG = -75.195;
+initial_coords = new google.maps.LatLng(INITIAL_CENTER_LAT, INITIAL_CENTER_LONG);
+
+// Initial map zoom value
+ZOOM = 17;
```
sandlerben: See comment above - I don't think hardcoding this will be necessary either

app/static/js/initialize_map.js
```diff
@@ -0,0 +1,119 @@
...
+
+// Bounds for the Date Slider. End date is current date.
+var BOUNDS_MIN = new Date(2015, 0, 1);
```
sandlerben: Let's make this date _really_ early (like 1980 or something) just in case

app/static/js/initialize_map.js
```diff
@@ -0,0 +1,119 @@
...
+var BOUNDS_MAX = new Date();
+
+function MarkerWrapper(actualMarker, incidentDate, contentString) {
+    this.actualMarker = actualMarker;
+    this.incidentDate = incidentDate;
+    this.contentString = contentString;
+}
+
+function storeMarkerState(markerWrappers, map, minDate) {
```
sandlerben: Do you want to comment what this function does?

app/static/js/initialize_map.js
```diff
@@ -0,0 +1,119 @@
...
+    globalMarkerWrappers = markerWrappers;
+    globalMap = map;
+    console.log(markerWrappers[0].actualMarker);
```
sandlerben: Was this for debugging?

app/static/js/initialize_map.js
```diff
@@ -0,0 +1,119 @@
...
+    for (mw = 0; mw < globalMarkerWrappers.length; mw++)
+    {
+        (globalMarkerWrappers[mw].actualMarker).setMap(globalMap);
```
sandlerben: Out of curiosity - how did you fix this bug?

app/static/js/initialize_map.js
```diff
@@ -0,0 +1,119 @@
...
+    BOUNDS_MIN = minDate;
+}
+
+// Use Google geocoder to update geolocation given an address through
+// the address search box
+function update_center() {
+    geocoder = new google.maps.Geocoder();
+    address = $("#address").val();
+    if (geocoder) {
+        console.log(address);
```
sandlerben: Did you intend to leave these `console.log`s in?

app/static/js/initialize_map.js
```diff
@@ -0,0 +1,119 @@
...
+        geocoder.geocode( { 'address': address},
+        function(results, status) {
+            if (status != google.maps.GeocoderStatus.OK) {
+                alert("Geocode not successful - " + status);
+            }
+            else if (status == google.maps.GeocoderStatus.ZERO_RESULTS) {
+                alert("Address not found!");
```
sandlerben: Can we do this in a more semantic-ui-ish type of way

http://semantic-ui.com/collections/message.html#negative--error

app/static/js/initialize_map.js
sandlerben: Actually ignore this comment because this date shows up in the dateslider

I think min date should be the min date of all the reports and max date should be the max date of all reports

app/static/styles/map.css
sandlerben: agreed^

app/static/js/initialize_map.js
sandlerben: Let's put this file in assets instead of static - that way it will be [minified](https://en.wikipedia.org/wiki/Minification_%28programming%29) and compressed

app/templates/main/map.html
```diff
@@ -0,0 +1,122 @@
+{% extends 'layouts/base.html' %}
+{% block content %}
+<link rel="stylesheet" href="static/styles/map.css" type="text/css" />
```
sandlerben: If you make `map.css` as a sass file and then move it to assets you won't have to import it here

app/templates/main/map.html
```diff
@@ -0,0 +1,122 @@
...
+<script src="static/js/initialize_map.js"></script>
```
sandlerben: same as above

app/static/js/initialize_map.js
sanjayss34: I changed the loop type from for-each type to a loop going through each index

app/static/js/initialize_map.js
sandlerben: You can find the min and max dates by querying with the `order_by` function and then getting the first element when you `order` ascendingly and then the first element when you `order` descendingly.

app/static/js/initialize_map.js
sandlerben: Interesting...

I didn't know javascript had for-each loops, maybe they work strangely

kyle-rosenbluth:
http://stackoverflow.com/questions/18444161/google-maps-responsive-resize

app/__init__.py
rileynwong: what might be cleaner would be `vendor.css` and `vendor.js` in their respective folders

rileynwong: `iThing` is also suuuuuper vague, pls rename to something more specific? what is it?

sanjayss34: iThing is from a vendor, so do you still want me to rename it?

sandlerben:
It looks like this PR is failing on circleci because of pep8 violations. Is flake8 working locally?

app/templates/main/map.html
```diff
@@ -0,0 +1,108 @@
+{% extends 'layouts/base.html' %}
+{% block content %}
+
+<script type="text/javascript">
+    function initializeMapAndMarkers() {
```
sandlerben, line 36: I would put all your code inside a 

``` javascript
$( document ).ready(function() {

});
```

Read why [here](https://learn.jquery.com/using-jquery-core/document-ready/)

app/templates/main/map.html
```diff
@@ -0,0 +1,108 @@
...
+        {% endfor %}
+        var mapProp = {
+                mapTypeId: google.maps.MapTypeId.ROADMAP,
+                mapTypeControl: false,
+                streetViewControl: false
+        };
+        map=new google.maps.Map($('#googleMap')[0],mapProp);
+
+        // Use HTML geolocation to center map if possible
+        if (navigator.geolocation) {
```
sandlerben: Do we want to do this by default or just when they select the geolocate button?

@kongsally

app/templates/main/map.html
```diff
@@ -0,0 +1,108 @@
...
+                };
+                map.setCenter(pos_center);
+            });
+        }
+        else {
+            console.log("Browser is not supporting geolocation.");
+        }
+        storeMarkerState(markerWrappers, map, minDate, bounds);
+        initializeDateSlider();
+        console.log("Done\n");
```
sandlerben: I would remove the `console.log`s which were just for your debugging

app/templates/main/map.html
```diff
@@ -0,0 +1,108 @@
...
+    }
+    // Add the map to the window
+    google.maps.event.addDomListener(window, 'load', initializeMapAndMarkers);
+
+    $(function () {
+        $("#googleMap").css("height", $(window).height());
```
sandlerben: Do you mind explaining what this does (leave a comment in the code)

app/templates/main/map.html
```diff
@@ -0,0 +1,108 @@
...
+    });
+</script>
+<div class ="column">
+    <div id="wrapper">
+            <div id="googleMap"></div>
+            <div id="slider"></div>
+            <form name = "addressForm" id = "addressForm">
```
sandlerben: nit: no space around = sign

app/assets/scripts/map.js
```diff
@@ -0,0 +1,121 @@
...
+        });
+    }
+}
+
+$('.close.icon').on('click', function() {
+    $(this).parent().hide();
+});
+
+// Get address submit event
+$(document).ready(function() {
```
sandlerben, line 153: I would actually put all the javascript code in this file inside this (I linked to an explanation why below)

sandlerben:
Looks pretty good except for a few minor things which I commented on. Once we get the modal working and tests passing, this should be good to merge.

@nancywong can you give this one final review?

sandlerben:
Oh also, we should revert the merge from the report-public-private branch. I gave instructions on how to do this on slack.

app/templates/main/map.html
sanjayss34: Doesn't the browser ask the user to use the user location anyway, so it's not really by default?

sandlerben: Doesn't this code check the user's location? Is that part of google maps by default?

sanjayss34: I believe that the only part of this code that checks the user's location is the "if(navigator.geolocation)" part, which is triggered only if the user allows her/his current location to be used.

app/assets/scripts/map.js
sanjayss34, line 153: I think that because there are references to functions in the code outside the file, this would cause problems related to scope (I got errors when I just tried it).

app/assets/styles/map.scss
```diff
@@ -0,0 +1,47 @@
+
+.ui.text.container {
+    text-align: center;
```
kongsally: What is this for? This affects the ui text container div in every page

sanjayss34:
@kongsally @sandlerben Is this branch ready to be merged with master?

app/main/views.py
```diff
@@ -1,6 +1,7 @@
 from flask import render_template
 from ..models import EditableHTML
 from . import main
+from app.models import IncidentReport
```
sandlerben: Can you put this in the import above?

``` python
from ..models import EditableHTML, IncidentReport
```

app/templates/main/map.html
sandlerben, line 36: bump

app/assets/scripts/map.js
sandlerben, line 153: Ok, you can keep functions outside but all listeners must be inside this

app/templates/main/map.html
```diff
@@ -0,0 +1,106 @@
...
+                var pos_center = {
+                    lat: position.coords.latitude,
+                    lng: position.coords.longitude
+                };
+                map.setCenter(pos_center);
+            });
+        }
+        storeMarkerState(markerWrappers, map, minDate, bounds);
+        initializeDateSlider();
+        console.log("Done\n");
```
sandlerben: Do you need this here?

app/templates/main/map.html
```diff
@@ -0,0 +1,106 @@
...
+    }
+    // Add the map to the window
+    google.maps.event.addDomListener(window, 'load', initializeMapAndMarkers);
+
+    // Set the height of the googleMap to the height of the window (less
```
sandlerben: This should be inside a `$( document ).ready`

app/templates/main/map.html
```diff
@@ -0,0 +1,106 @@
...
+    // any elements above the googleMap)
+    $(function () {
```
sandlerben: This should be inside a `$( document ).ready`

app/templates/main/map.html
```diff
@@ -0,0 +1,106 @@
...
+                };
+                map.setCenter(pos_center);
+            });
+        }
+        storeMarkerState(markerWrappers, map, minDate, bounds);
+        initializeDateSlider();
+        console.log("Done\n");
+    }
+    // Add the map to the window
+    google.maps.event.addDomListener(window, 'load', initializeMapAndMarkers);
```
sandlerben: This should be inside a `$( document ).ready`

app/templates/main/map.html
sandlerben, line 36: Actually, only listeners should be inside

app/assets/scripts/map.js
```diff
@@ -0,0 +1,121 @@
...
+                globalMap.setCenter(results[0].geometry.location);
+            }
+            $('.close.icon').on('click', function() {
+                $(this).parent().hide();
+            });
+        });
+    }
+}
+
+$('.close.icon').on('click', function() {
```
sandlerben: Put this inside the `$(document).ready`

app/templates/main/map.html
sandlerben: I don't think that's correct

https://developer.mozilla.org/en-US/docs/Web/API/Navigator/geolocation

I think we should have a button (probably the default google maps one will work) to ask for current location.

app/main/views.py
```diff
@@ -8,6 +9,15 @@ def index():
     return render_template('main/index.html')
 
 
+@main.route('/map')
```
sandlerben, line 7: You should combine this with the index ('/') route. 

@kongsally we don't need the current index route do we?

sandlerben:
Imgur

I feel like there is a bit too much padding between the button and the search box. Also, the search box is round while the button is more rectangular. Are we sure we want to make the button gray?

@kongsally what do you think?

app/templates/main/map.html
sanjayss34: As I commented above, putting all of the code in this file in a $(document).ready causes scope errors

app/templates/main/map.html
sanjayss34: As the web page you linked says, "For security reasons, when a web page tries to access location information, the user is notified and asked to grant permission. Be aware that each browser has its own policies and methods for requesting this permission." I'm not sure then why you would like a button; it seems like the user would have to press both "Allow" (for the navigator.geolocation to be used) and whatever button we create for a single purpose. Could you please explain what I am missing?

sanjayss34:
^: I can decrease the padding between the button and the search box, make the button the teal color of the menu bar, and try to make the button round.

app/templates/main/map.html
sandlerben: Sure. I just opened up the page on my laptop and the map immediately shifted to my location. What if I just wanted to see all the pins? I think the user should have some control over when we request for their location.

We can do this in another PR, though. [Here](https://jsfiddle.net/mi3afzal/ogsvzacz/2/) is code that essentially does it.

sandlerben:
Before we merge, I want to verify that this looks right on smaller screens. I'm going to deploy a heroku review app for this PR.

sanjayss34:
I don't understand the purpose of the geolocation button. 1) I would think that the default should be using current location so that users will find it easier to locate places/incidents close to them, and 2) the user has to click both the new geolocation button and the allow to use location button to use current location. Why is this approach advantageous?

app/templates/main/map.html
```diff
@@ -0,0 +1,100 @@
...
+        initializeDateSlider();
+        addLocationButton(map);
+    }
+    $(document).ready(function() {
+        // Add the map to the window
+        google.maps.event.addDomListener(window, 'load', initializeMapAndMarkers);
+
+        // Set the height of the googleMap to the height of the window (less
+        // any elements above the googleMap)
+        $(function () {
```
kongsally, line 116: You might want to change the margin top of the footer here so that we don't have white space between the map and the footer

app/assets/styles/map.scss
```diff
@@ -0,0 +1,56 @@
+@import "app";
+
+.ui-rangeSlider-bar {
+    background: $dark-teal;
```
kongsally: change it to $light-teal

app/assets/styles/map.scss
```diff
@@ -0,0 +1,56 @@
...
+}
+
+#addressSearch {
+    position: absolute;
+    left: 30px;
+    top: 49px;
+    z-index: 99;
+}
+#submitButton {
+    background-color: $dark-teal;
```
kongsally: change this to $light-teal as well :)

sandlerben:
Sanjay we can chat about geolocation logic over slack or in person. I think this flow is preferable but we can easily change it back.

PR #3, opened 2015-10-12
Add basic idling incident schema

kyle-rosenbluth:
Some things to discuss:
- How do we want location represented? Here I have it as a string, but I could also see it being a set of lat,lon coordinates.
- Should `picture` be a string (url) or data?

cc @sandlerben

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -192,26 +192,25 @@ def load_user(user_id):
 class Location(db.Model):
     __tablename__ = 'locations'
     id = db.Column(db.Integer, primary_key=True)
-    latitude = db.Column(db.String(50))
-    longitude = db.Column(db.String(50))
+    latitude = db.Column(db.String(50), nullable=False)
+    longitude = db.Column(db.String(50), nullable=False)
     original_text = db.Column(db.Text)
+    idling_incident_id = db.Column(db.Integer,
+                                   db.ForeignKey('idling_incidents.id'))
 
 
 class IdlingIncident(db.Model):
-    __tablename__ = 'incidents'
+    __tablename__ = 'idling_incidents'
     id = db.Column(db.Integer, primary_key=True)
-    vehicle_id = db.Column(db.Integer)
-    location = db.relationship("Location", uselist=False,
-                               backref="idling_incident")
-    date = db.Column(db.DateTime)
-    agency = db.Column(db.String(64))
-    picture = db.Column(db.String(64))
+    vehicle_id = db.Column(db.String(8))
+    license_plate = db.Column(db.String(8))
+    location = db.relationship('Location',
+                               uselist=False,
+                               lazy='joined',
+                               backref='idling_incident')
+    date = db.Column(db.DateTime, nullable=False)  # hour the incident occurred
+    duration = db.Column(db.Interval, nullable=False)  # like timedelta object
+    # agency = TODO: Hook this up with agency when Nancy finishes. TODO: set
+    # nullable=False for this field, default=<Other>
+    picture_url = db.Column(db.Text)
     description = db.Column(db.Text)
-
-    def __init__(self, vehicle_id, location, date, agency, picture, description):
-        self.vehicle_id = vehicle_id
-        self.location = location
-        self.date = date
-        self.agency = agency
-        self.picture = picture
-        self.description = description
--- a/app/models.py
+++ b/app/models.py
@@ -194,7 +194,7 @@ class Location(db.Model):
     id = db.Column(db.Integer, primary_key=True)
     latitude = db.Column(db.String(50))
     longitude = db.Column(db.String(50))
-    original_text = db.Column(db.Text)
+    original_user_text = db.Column(db.Text)  # the raw text which we geocoded
     idling_incident_id = db.Column(db.Integer,
                                    db.ForeignKey('idling_incidents.id'))
 
--- a/app/models.py
+++ b/app/models.py
@@ -43,6 +43,22 @@ class Role(db.Model):
     def __repr__(self):
         return '<Role \'%s\'>' % self.name
 
+class IdlingIncident(db.Model):
+    __tablename__ = "incidents"
+    vehicle_id = db.Column(db.Integer, primary_key=True)
+    location = db.Column(db.String(64))
+    date = db.Column(db.DateTime)
+    agency = db.Column(db.String(64))
+    picture = db.Column(db.String(64))
+    description = db.Column(db.Text)
+
+    def __init__(self, vehicle_id, location, date, agency, picture, description):
+        self.vehicle_id = vehicle_id
+        self.location = location
+        self.date = date
+        self.agency = agency
+        self.picture = picture
+        self.description = description
 
 class User(UserMixin, db.Model):
     __tablename__ = 'users'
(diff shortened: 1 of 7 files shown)


sandlerben:
Thanks for the PR, will review before or during meeting tonight.

sandlerben:
Sorry that I haven't reviewed yet, I am planning to look during meeting.

For all model schema's we should add some basic tests to /tests

app/models.py
```diff
@@ -43,6 +43,22 @@ def insert_roles():
     def __repr__(self):
         return '<Role \'%s\'>' % self.name
 
+class IdlingIncident(db.Model):
+    __tablename__ = "incidents"
+    vehicle_id = db.Column(db.Integer, primary_key=True)
```
sandlerben: What is a `vehicle_id`? Is it like `vehicle_id` of 1 is SEPTA, 2 is something else, etc? If so, I think we should create a `VehicleType` model and have a relationship between  `IdlingIncident` and `VehicleType`. `IdlingIncident` can have an integer id (this can be the primary key) as well as a string name.

app/models.py
```diff
@@ -43,6 +43,22 @@ def insert_roles():
...
+    location = db.Column(db.String(64))
```
sandlerben: I would move location into its own data type. It would be somewhat like [this](https://github.com/hack4impact/gpcah/blob/master/app/models.py#L80) except that we only need `latitude`, `longitude` and `address`. `address` is just the string that the user submitted (which we attempt to geocode). It makes sense to keep this string in case we want to a) display it to the user or b) the geocode fails.

app/models.py
sandlerben: Wait, just saw that `vehicle_id` is different than `agency`. Can you explain the difference?

app/models.py
```diff
@@ -43,6 +43,22 @@ def insert_roles():
...
+    date = db.Column(db.DateTime)
+    agency = db.Column(db.String(64))
+    picture = db.Column(db.String(64))
```
sandlerben: To answer your question, this should just store a URL. I don't think we want to host photos ourselves.

sandlerben: @nancywong has some experience hosting images on S3. I also think there are some free photo hosting APIs we could check out.

app/models.py
```diff
@@ -43,6 +43,22 @@ def insert_roles():
     def __repr__(self):
         return '<Role \'%s\'>' % self.name
 
+class IdlingIncident(db.Model):
+    __tablename__ = "incidents"
```
sandlerben: Add an `id` primary key and make `vehicle_id` not the primary_key.

Try adding this line:

``` python
id = db.Column(db.Integer, primary_key=True)
```

app/models.py
sandlerben: Ok, so `vehicle_id` is just something government vehicles have. I think that we might just want to ask for license plate instead. @kongsally, can you ask them how they feel about this?

I think license plate is better because
1. Users already know where to find the plate on the car.
2. Agencies can map license plate to vehicle id.
3. We can ask consistent questions in the reporting form (e.g. we do not need to ask if this was a government vehicle)

app/models.py
sandlerben: Are we sure that the picture URL will only be 64 characters? Should it be required?

app/models.py
```diff
@@ -43,6 +43,22 @@ def insert_roles():
...
+    vehicle_id = db.Column(db.Integer, primary_key=True)
+    location = db.Column(db.String(64))
+    date = db.Column(db.DateTime)
+    agency = db.Column(db.String(64))
+    picture = db.Column(db.String(64))
+    description = db.Column(db.Text)
+
+    def __init__(self, vehicle_id, location, date, agency, picture, description):
```
sandlerben: This init assumes that each of these fields is required. If this is the case, we should explicitly make them required using `nullable=False` (see https://github.com/hack4impact/gpcah/blob/0e41db2c4da1964bc84cb81afedca972ef87831a/app/models.py#L285)

sandlerben:
Overall, this looks pretty good. Great work :+1: 

I have a few general questions. Do we want to make any of the fields required (e.g. description, location, picture, etc)? Are we sure that 64 characters is enough space for all the fields?

app/models.py
thomaslee1911: I think it's still valuable to have both (i.e. include another `license_plate` field).
1. This is just from me scrolling through `poll244.csv`  but there are many rows where they found the vehicle id but not license plate, e.g. rows 145, 757 to 759, 763 to 766. 
2. We should ask them about it; I'd imagine otherwise they wouldn't have asked in the past. Also above situation happens. 
3. I don't think CAC currently asks for whether it's government vehicle, and I don't think vehicle id is only for government ones (almost all observation rows have both).

sandlerben: @kongsally can you ask them about this?

thomaslee1911:
Also I think we should include a field for the duration, i.e. "How many minutes have you observed the vehicle idling? Stop your timer! Be specific!". It seems like the idling duration is needed to estimate total emissions (https://en.wikipedia.org/wiki/Idle_(engine) has pollution rates in g/min).

sandlerben:
@kyle-rosenbluth I'm going to commandeer this PR because it's blocking people

sandlerben:
From @kongsally 

For the Incident Schema it has the field Date and Time. Is time here the time when the incident happened or is it the time/duration of the idling incident? I think we would need both the time it happened and the duration of the idling (it’s illegal once its >5minutes)

sandlerben:
This isn't totally done, and I'd like to merge this and #6 at the same time. We can make one branch with both of the changes and then merge.

r @nancywong @kyle-rosenbluth @thomaslee1911

kyle-rosenbluth:
Looks good to me

app/models.py
```diff
@@ -187,3 +187,30 @@ def is_admin(self):
...
 def load_user(user_id):
     return User.query.get(int(user_id))
+
+
+class Location(db.Model):
+    __tablename__ = 'locations'
+    id = db.Column(db.Integer, primary_key=True)
+    latitude = db.Column(db.String(50))
+    longitude = db.Column(db.String(50))
+    original_text = db.Column(db.Text)
```
rileynwong: what's the `original_text` field for? unclear - maybe rename variable or add a line comment?

sandlerben: It is the raw text the user imports. I will rename it `original_user_text`.

kyle-rosenbluth: what about `user_input`

app/models/idling_incident.py
```diff
@@ -0,0 +1,28 @@
...
+    location = db.relationship('Location',
+                               uselist=False,
+                               lazy='joined',
+                               backref='idling_incident')
+    date = db.Column(db.DateTime)  # hour the incident occurred
+    duration = db.Column(db.Interval)  # like timedelta object
+    # agency = TODO: Hook this up with agency when Nancy finishes. TODO: set
+    # default=<Other> for this field
+    picture_url = db.Column(db.Text)
+    description = db.Column(db.Text)
```
thomaslee1911: We need the agency name too right?

sandlerben:
Closing because these changes will be merged with #6

PR #4, opened 2015-10-16
Refactored models from file to a directory

sandlerben:


sandlerben:
r @nancywong 

Also, @aharelick or @jondubin can you tell me if you think this is the right approach? Should we do it this way in flask-base?

app/models/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""
+These imports enable us to make all defined models members of the models
+module (as apposed to just their python files)
```
rileynwong: typo - opposed

sandlerben: :panda_face:

maxmcc:
Jumping in here: I think this is a _much_ better structure than we've had in previous projects. We should definitely adopt this in flask-base.

sandlerben:
Merged into flask-base instead

PR #5, opened 2015-10-17
Added phone number field to user

sandlerben:


sandlerben:
r @thomaslee1911 @kongsally

app/account/forms.py
```diff
@@ -4,10 +4,10 @@
     StringField,
     PasswordField,
     BooleanField,
-    SubmitField
+    SubmitField,
```
thomaslee1911, line 6: Do we need this comma?

app/account/forms.py
```diff
@@ -37,6 +37,12 @@ class RegistrationForm(Form):
         Length(1, 64),
         Email()
     ])
+    phone_number = TelField('Phone Number', validators=[
+        Length(0, 15),
```
thomaslee1911: Should the minimum length be 10?

app/account/forms.py
```diff
@@ -37,6 +37,12 @@ class RegistrationForm(Form):
...
+        Regexp(r'^[0-9]*$',
```
thomaslee1911: We discussed this a little during meeting, I think we said we'd consider allowing symbols but removing them later?

app/account/forms.py
sandlerben, line 6: Technically no, but it's good style to leave an extra comma in situations like this

For example, Go won't even compile if you leave out the dangling comma!

app/account/forms.py
sandlerben: Apparently for international numbers, 15 is the most it can be.

app/account/forms.py
sandlerben: Yes, going to do that!

sandlerben:
Now all phone number formats are accepted. The non-digit characters are stripped out.

Look good now @kongsally @thomaslee1911 ?

sandlerben:
Ok, just waiting for a :+1: from @thomaslee1911

PR #6, opened 2015-10-17
Create agency schema

rileynwong:


rileynwong:
r @thomaslee1911

app/models.py
```diff
@@ -8,6 +8,7 @@
 
 class Permission:
     GENERAL = 0x01
+    WORKER = 0x10
```
sandlerben: Can we rename this to `AGENCYWORKER`? In case there are other types of workers which will be added in the future.

app/models.py
```diff
@@ -26,6 +27,9 @@ def insert_roles():
             'User': (
                 Permission.GENERAL, 'main', True
             ),
+            'AgencyWorker': (
+                Permission.WORKER, 'main', True
```
sandlerben: Do we not want to give `AgencyWorker` it's own blueprint? Just like `Administrator` has it's `admin` blueprint?

sandlerben: With regards to `True` (aka the third element in the tuple): what does this mean? I know it sets the `default` field to `True`. What does this mean?

I hypothesize that it means that users are, by default, given the role with `default` of True (e.g. right now they are given User role). Thus, if there are several roles with `default` of True, this won't work.

Can you check how it works?

app/models.py
```diff
@@ -186,3 +191,35 @@ def is_admin(self):
 @login_manager.user_loader
 def load_user(user_id):
     return User.query.get(int(user_id))
+
+
+# Agency schema
```
sandlerben: Nit: if you are going to comment about a class use docstrings like this http://stackoverflow.com/questions/3051241/how-to-document-class-attributes-in-python?rq=1

app/models.py
```diff
@@ -186,3 +191,35 @@ def is_admin(self):
...
+class Agency(db.Model):
+    __tablename__ = 'agencies'
+    id = db.Column(db.Integer, primary_key=True)
+    name = db.Column(db.String(128), index=True)
```
sandlerben: Should name be `unique=True`?

app/models.py
```diff
@@ -186,3 +191,35 @@ def is_admin(self):
...
+    is_public = db.Column(db.Boolean, default=False)
+    users = db.relationship('User', backref='agency', lazy='dynamic')
+    reports = db.relationship('IncidentReport', backref='agency',
```
sandlerben: For this line and the one above it, why are you using `lazy='dynamic'`?

According to docs:

> dynamic - the attribute will return a pre-configured Query object for all read operations, onto which further filtering operations can be applied before iterating the results. See the section Dynamic Relationship Loaders for more details.

Do you really think we are going to do more filtering when we get all the users or all reports for a `Agency`? Do you think we will need filtering when we get all reports for a user or a agency? 

We may want to start with `lazy='select'` or `lazy='joined'` and then change it if we need to.

app/models.py
```diff
@@ -186,3 +191,35 @@ def is_admin(self):
...
+                              lazy='dynamic')
+
+    @staticmethod
+    def insert_agencies():
+        agencies = {
+            'SEPTA': (
+                False
```
sandlerben: @kongsally can you give Nancy a list of the agencies we want by default?

sandlerben:
@nancywong looks pretty good, I made a few comments.

sandlerben:
r @nancywong @kyle-rosenbluth @sanjayss34

sandlerben:
The parts I didn't write look good to me!

app/models/agency.py
```diff
@@ -0,0 +1,50 @@
...
+class Agency(db.Model):
+    __tablename__ = 'agencies'
+    id = db.Column(db.Integer, primary_key=True)
+    name = db.Column(db.String(128), index=True, unique=True)
+
+    # True if this agency was officially created by an Administrator. Official
+    # agencies can have relationships to AgencyWorkers and will show up in the
+    # reporting form. Non-official agencies will not show up in the form.
+    is_official = db.Column(db.Boolean, default=False)
+    is_public = db.Column(db.Boolean, default=False)
```
rileynwong, line 17: Could we add a comment to clarify `is_public` as well? Otherwise lgtm

sandlerben, line 17: Good idea

sanjayss34:
Hey I'm not really understanding the Role.insert_roles() and Agency.insert_agencies() functions. Could someone explain them to me? In particular, what are the structures that have parentheses around them and inside have key-like things within single quotes, e.g. 'User' in insert_roles(), and 'SEPTA' in insert_agencies(). Thanks!

sandlerben:
@sanjayss34 I can walk you through the Python syntax in-person and explain how that works.

The idea of the `insert` functions are that there are a set of `Role` and `Agency` which a `User` can have a relationship with. When we first run the app, we have to add these default `Role` and `Agency` objects to the database (e.g. Administrator). 

If you are familiar with enums, you can think of `Role` and `Agency` as a dynamic set of enums.

sanjayss34:
Thanks @sandlerben . Since we have to add the default Role objects to the database when we first run the app, how come there is not db.session.commit() call in insert_roles() as there is in insert_agencies()? Sorry if this is a stupid question!

aharelick:
Looks like there is one right here: https://github.com/hack4impact/clean-air-council/blob/master/app/models/user.py#L41

sanjayss34:
Okay, got it, thanks. The code looks good to me.

rileynwong:
:shipit:

kyle-rosenbluth:
LGTM

PR #7, opened 2015-10-24
User reports relationship

kyle-rosenbluth:
Do we need tests for this?
@sandlerben


sandlerben:
Please incorporate master so I can better tell what you added vs what was added recently in master

app/models/user.py
```diff
@@ -62,6 +62,7 @@ class User(UserMixin, db.Model):
     password_hash = db.Column(db.String(128))
     role_id = db.Column(db.Integer, db.ForeignKey('roles.id'))
     agency_id = db.Column(db.Integer, db.ForeignKey('agencies.id'))
+    reported_incidents = db.Relationship("IdlingIncident")
```
sandlerben: Can we make the relationship in the IdlingIncident model just to be consistent with how we made other relationships to User?

sandlerben: Also, when creating a relationship make sure to pass arguments like `backref` and `lazy`!

sandlerben:
Please test!

kyle-rosenbluth:
Can we talk about how we would test something like this? Not quite sure what to do.

sandlerben:
I would add one test to `test_user_model.py` and one test to `test_incident_report.py`. You can model them after a test case like `test_incident_report_with_agency_no_location`

kyle-rosenbluth:
Ready for final review @sandlerben

sandlerben:
:shipit:

PR #14, opened 2015-10-30
Add default accounts for admin, agency worker, and general users

rileynwong:


manage.py
```diff
@@ -65,18 +65,38 @@ def add_fake_data(number_users, number_reports):
     User.generate_fake(count=number_users)
     IncidentReport.generate_fake(count=number_reports)
 
+
```
sandlerben, line 61: Why the added space here? Was it not there before?

manage.py
```diff
@@ -65,18 +65,38 @@ def add_fake_data(number_users, number_reports):
...
 @manager.command
 def setup_dev():
     """Runs the set-up needed for local development."""
     setup_general()
 
     # Create a default admin user
-    admin = User(email='user@example.com',
+    admin = User(email='user@admin.com',
```
sandlerben: Why don't we make the emails like `admin@user.com`?

sandlerben:
:shipit:

manage.py
rileynwong, line 61: yea pep8 was >:( at me

sandlerben, line 61: Yay pep8

PR #10, opened 2015-10-24
List reports 

rileynwong:
Endpoint is reports/reports

PR #20, opened 2015-10-30
View list of incident reports

rileynwong:


rileynwong:
r @kyle-rosenbluth ?

sandlerben:
Can we close #10?

app/reports/views.py
```diff
@@ -0,0 +1,31 @@
...
+from flask.ext.login import login_required, current_user
+from . import reports
+from ..models import IncidentReport, Agency
+
+
+@reports.route('/all')
+@login_required
+def view_reports():
+    """View all idling incident reports."""
+    """Different roles have access to different reports.
```
sandlerben: Nit: can we combine this into one docstring?

app/reports/views.py
```diff
@@ -0,0 +1,31 @@
...
+    A general user can see all reports submitted by that user."""
+
+    if current_user.is_admin():
+        reports = IncidentReport.query.all()
+        agencies = Agency.query.all()
+
+    elif current_user.is_agency_worker():
+        agency_id = current_user.agency.id
+        agencies = Agency.query.filter_by(id=agency_id)
+        reports = IncidentReport.query.filter_by(agency_id=agency_id)
```
sandlerben: Couldn't we just do

`current_user.agency.reports`?

app/templates/reports/reports.html
```diff
@@ -0,0 +1,106 @@
...
+                </div>
+            </h2>
+
+            <div class="ui menu">
+                <div id="select-agency" class="ui dropdown item">
+                    <div class="text">
+                        {% if current_user.is_admin() %}
+                            All incident reports
+                        {% else %}
+                            {{ current_user.agency.name }}
```
sandlerben: Maybe change to `{{ current_user.agency.name }}'s incident reports`

sandlerben:
I'll review a bit more closely when this is all finished up (e.g. logic for general users). How do you think users should get to this page? Should we add an `reports` link to the navbar?

rileynwong:
@sandlerben @thomaslee1911

app/reports/views.py
```diff
@@ -0,0 +1,30 @@
...
+    An Agency worker can all reports for their agency.
+    A general user can see all reports submitted by that user."""
+
+    if current_user.is_admin():
+        reports = IncidentReport.query.all()
+        agencies = Agency.query.all()
+
+    elif current_user.is_agency_worker():
+        reports = current_user.agency.reports
+        agencies = None
```
thomaslee1911: Should the agency worker be associated with their agency here?

app/models/incident_report.py
```diff
@@ -11,6 +11,9 @@ class Location(db.Model):
     incident_report_id = db.Column(db.Integer,
                                    db.ForeignKey('incident_reports.id'))
 
+    def __repr__(self):
+        return '<Location \'%s\'>' % self.longitude
```
thomaslee1911: Not sure where this is used but maybe include latitude as well.

app/reports/__init__.py
```diff
@@ -0,0 +1,5 @@
+from flask import Blueprint
+
+reports = Blueprint('reports', __name__)
+
+from . import views  # noqa
```
thomaslee1911, line 5: Could you do this in accordance with flake8 / without using noqa?

rileynwong, line 5: no, i think this is what we do for any `__init__.py` file

app/models/incident_report.py
rileynwong: good call, i changed this to just return lat, long in string format

app/models/incident_report.py
```diff
@@ -11,6 +11,9 @@ class Location(db.Model):
     incident_report_id = db.Column(db.Integer,
                                    db.ForeignKey('incident_reports.id'))
 
+    def __repr__(self):
+        return '%s, %s' % (self.latitude, self.longitude)
```
sandlerben: Nit: Let's use string.format

http://stackoverflow.com/questions/5082452/python-string-formatting-vs-format

app/models/incident_report.py
```diff
@@ -61,6 +65,7 @@ def flip_coin():
                 date=fake.date_time_between(start_date="-1y", end_date="now"),
                 duration=timedelta(minutes=randint(1, 30)),
                 agency=choice(agencies),
+                user=choice(users),
```
sandlerben, line 69: So every report will have a user but it's possible not every user will have a report? Sounds good.

app/reports/__init__.py
sandlerben, line 5: she is right

app/reports/views.py
```diff
@@ -0,0 +1,32 @@
+from flask import render_template
+from flask.ext.login import login_required, current_user
+from . import reports
+from ..models import IncidentReport, Agency
+
+
+@reports.route('/all')
+@login_required
+def view_reports():
+    """View all idling incident reports.
```
sandlerben, line 10: Nice docstring :+1:

app/reports/views.py
```diff
@@ -0,0 +1,32 @@
...
+
+    if current_user.is_admin():
+        reports = IncidentReport.query.all()
+        agencies = Agency.query.all()
+
+    elif current_user.is_agency_worker():
+        reports = []
+        for agency in current_user.agencies:
+            reports.extend(agency.incident_reports)
+        agencies = None
```
sandlerben: I would set agencies to None before the if/elif/else block. This takes advantage of what I consider to be a confusing and peculiar feature of python.

sandlerben:
A few comments/questions on the page itself:
1. What does `Submitter` mean?
2. Can we format the date nicer?
3. Does it really make sense to have a location column?
4. There is a funky bug with the coloring of the switch.
5. When I select SEPTA and then All incident reports I see no reports in the table

sandlerben:
@nancywong I'm seeing the warning "This pull request is closed, but the list-reports-b branch has unmerged commits."

Are all those unmerged commits on your new branch?

rileynwong:
yes, everything's on #57, which was branched off of this w/ the new commits

PR #21, opened 2015-10-30
Removed logic for adding users

sandlerben:
Admins only need to invite users


sandlerben:
r @apiatski @kongsally

rileynwong:
wait how come? isn't direct add easier and faster?

PR #22, opened 2015-10-30
Update default users

rileynwong:


sandlerben:
Nit: you might want to squash the two identical commits.

Besides that, if it works locally for you feel free to merge

PR #23, opened 2015-10-30
Added fake lat and long

sandlerben:


sandlerben:
r @sanjayss34 @apiatski

sanjayss34:
Code looks good!

PR #25, opened 2015-10-31
Parse CSV

thomaslee1911:
Script to parse existing poll244.csv file into agency, location, and incident report tables.


app/utils.py
```diff
@@ -29,3 +30,54 @@ def parse_phone_number(phone_number):
...
+    city_default = ', philadelphia, pennsylvania, usa'
+    vehicle_id_index = 8
+    license_plate_index = 9
+    location_index = 4
+    date_index = 0
+    agency_index = 6
+    picture_index = 13
+    description_index = 11
+    with open(filename, 'rb') as file:
+        reader = csv.reader(file, delimiter=',')
```
sandlerben: Why are you specifying a `delimiter`?

app/utils.py
```diff
@@ -29,3 +30,54 @@ def parse_phone_number(phone_number):
         stripped = '1' + stripped
     stripped = '+' + stripped
     return stripped
+
+
+def parse_to_db(db, filename):
+    city_default = ', philadelphia, pennsylvania, usa'
```
sandlerben: This is a bit of a hack. The Google geocoding API has [Viewport biasing](https://developers.google.com/maps/documentation/geocoding/intro#Viewports) which would be a more robust way of doing this.

sandlerben: I would check if `geocoder` supports this and if it doesn't we may have to use the geocoding api directly :(

app/utils.py
```diff
@@ -29,3 +30,54 @@ def parse_phone_number(phone_number):
...
+            start_time = datetime.datetime.strptime(row[date_index], date_format)
+            end_time = datetime.datetime.strptime(row[date_index + 1], date_format)
+            # Assign correct agency id
+            agency_name = row[agency_index].rstrip().upper()
+            if agency_name == 'OTHER':
+                agency_name = row[agency_index + 1].rstrip().upper()
+            a = Agency.query.filter_by(name=agency_name).first()
+            if a is None:
+                a = Agency(name=agency_name)
+                a.is_public = False
```
sandlerben: I think for `other` agencies public should be True by default.

@kongsally ?

app/utils.py
```diff
@@ -29,3 +30,54 @@ def parse_phone_number(phone_number):
...
+                a.is_official = False
+                db.session.add(a)
+                db.session.commit()
+            incident = IncidentReport(
+                vehicle_id=row[vehicle_id_index],
+                license_plate=row[license_plate_index],
```
sandlerben: Are there any cells which are blank for `vehicle_id` or `license_plate`? If so, I would put `None` here instead of empty string to be consistent.

poll244_sample.csv
```diff
@@ -0,0 +1,4 @@
+Timestamp:first,Timestamp:last,Are you ready to report an idling vehicle?,Are you ready to report an idling vehicle? (option value),"What is your location? Be specific! If you have a timer, press start!",What agency owns the vehicle you see idling?,What agency owns the vehicle you see idling? (option value),What agency or company?,What is the vehicle ID #? Usually on the back or side of the vehicle.  ,What is the license plate number? If you don't see it that's ok.,How many minutes have you observed the vehicle idling? Stop your timer! Be specific!,Please let us know of the general situation (i.e. is the vehicle unoccupied).,"Lastly, can you take a photo of the vehicle and text it back to us? Make sure plate or ID # is visible.","Lastly, can you take a photo of the vehicle and text it back to us? Make sure plate or ID # is visible. (media url)",Area Code,Phone Prefix
```
sandlerben: Did you mean to commit this file?

sandlerben:
Looks pretty good, once you or I make the changes we can have another round of reviews

poll244_sample.csv
thomaslee1911: I meant to commit it because the tests use it.

app/utils.py
thomaslee1911: I don't see it in the [documentation here](https://pypi.python.org/pypi/geocoder), so we can definitely use geocoding directly which shouldn't be hard.

My thought was that this parser is only for this particular csv format for the historical data (which would only be from Philly)? Also the viewport is a rectangular bounding box while Philadelphia's city boundaries are oblongish and not rectangular, so there's a robustness tradeoff there too?

thomaslee1911:
Thanks for the review. I can make the changes.

app/utils.py
kongsally: Yup! Other agencies should be public/true by default.

app/utils.py
thomaslee1911: Oops it's redundant.

app/utils.py
thomaslee1911: Got it thanks.

app/utils.py
thomaslee1911: Yes there are blank cells. I'll use None for those instead.

thomaslee1911: There are also cells like "Didn't see it", "Na", or string "None". I'll make this fix - keep those as strings, and have `None` for blanks?

sandlerben:
Looks good, just change it to use viewport biasing

app/utils.py
```diff
@@ -29,3 +33,61 @@ def parse_phone_number(phone_number):
         stripped = '1' + stripped
     stripped = '+' + stripped
     return stripped
+
+
+def parse_to_db(db, filename):
+    viewport_default = '39.861204,-75.310357|40.138932,-74.928582'
```
sandlerben: Why do we give it two sets of coordinates?

Can we add some attribute in config that contains the viewport default? That way if we redeploy the app for other cities, we can change that config and all will still work.

sandlerben: I now understand why we give it two sets of coordinates.

sandlerben: Actually, this is fine for now

app/utils.py
```diff
@@ -29,3 +33,61 @@ def parse_phone_number(phone_number):
...
+    with open(filename, 'rb') as file:
+        reader = csv.reader(file)
+        columns = reader.next()
+        for row in reader:
+            address_text = row[location_index]
+            # Viewport-biased geocoding using Google API
+            url = "https://maps.googleapis.com/maps/api/geocode/json"
+            payload = {'address': address_text, 'bounds': viewport_default}
+            r = requests.get(url, params=payload)
+            coordinates = r.json()['results'][0]['geometry']['location']
```
sandlerben: What happens if it fails to geocode?

app/utils.py
```diff
@@ -29,3 +33,61 @@ def parse_phone_number(phone_number):
...
+                original_user_text=address_text)
+            db.session.add(loc)
+            date_format = "%m/%d/%Y %H:%M"
+            time1 = datetime.datetime.strptime(row[date_index], date_format)
+            time2 = datetime.datetime.strptime(row[date_index+1], date_format)
+            # Assign correct agency id
+            agency_name = row[agency_index].rstrip().upper()
+            if agency_name == 'OTHER':
+                agency_name = row[agency_index + 1].rstrip().upper()
+            a = Agency.query.filter_by(name=agency_name).first()
```
sandlerben: I'm going to add a get_agency_by_name method that ensures this query is case-sensitive. When I'm done, we can change this.

app/utils.py
```diff
@@ -29,3 +33,61 @@ def parse_phone_number(phone_number):
...
+            if a is None:
+                a = Agency(name=agency_name)
```
sandlerben: Make sure agency_name is all caps

thomaslee1911: This should be done by line 66 or 68, but we can change to your case-insensitive method.

thomaslee1911:
I've moved the viewport coordinates to config and now use Ben's get_agency_by_name changes to ensure case-insensitivity.

poll244_sample.csv
sandlerben: sounds good

config.py
```diff
@@ -20,6 +20,7 @@ class Config:
     EMAIL_SUBJECT_PREFIX = '[{}]'.format(APP_NAME)
     EMAIL_SENDER = '{app_name} Admin <{email}>'.format(app_name=APP_NAME,
                                                        email=MAIL_USERNAME)
+    VIEWPORT = '39.861204,-75.310357|40.138932,-74.928582'
```
sandlerben, line 26: Maybe name this something a bit more semantic and explain what it is used for?

sandlerben:
Looks good, once we know how many reports are not successfully geocoded and what to do with them we can merge

poll244_sample.csv
sandlerben: Let's put this in a tests folder then

sandlerben:
What is tests_backup?

sandlerben:
@thomaslee1911 How do I run the script locally and see how many fail to geocode?

thomaslee1911:
I moved the actual tests to test_backup, so test only has the script that parses the entire file. To run the script you should be able to just `python manage.py test`. It should output a count of geocode no-responses and a string with each of the failed addresses.

sandlerben:
That seems like a hack^ Can we add a manage.py argument so we could run it like

`python manage.py parse_csv --file "<file name here">

sandlerben:
At this point, the code looks pretty good. However, I figure that there are quite a few edge cases we may not be considering, so I will wait until the rest of the app is done before we merge (that way we can do a complete test).

PR #29, opened 2015-11-07
Twilio integration

kyle-rosenbluth:


sandlerben:
@kyle-rosenbluth do you want me to review?

kyle-rosenbluth:
Not yet, I'll ping you later tonight when it's ready

sandlerben:
r @nancywong @kyle-rosenbluth @kongsally

app/assets/scripts/map.js
```diff
@@ -157,6 +157,40 @@ $(document).ready(function() {
     });
 });
 
+function getNextDate(startDate) {
+    date = startDate.getDate();
+    month = startDate.getMonth();
+    year = startDate.getFullYear();
+    MAX_MONTH = 11;
```
rileynwong: Why 0-11 and not 1-12? I think 1-12 would make more sense. In any case, should add a docstring here imo.

app/assets/scripts/map.js
```diff
@@ -157,6 +157,40 @@ $(document).ready(function() {
...
+        4: 31,
+        5: 30,
+        6: 31,
+        7: 31,
+        8: 30,
+        9: 31,
+        10: 30,
+        11: 31
+    }
+    if ((month == 1) && ((year % 400 == 0) || ((year % 100 != 0) && (year % 4 == 0)))) {
```
rileynwong: I would factor this out into a helper fxn, maybe call it `isLeapYear` that returns a boolean

app/main/messaging.py
```diff
@@ -0,0 +1,392 @@
...
+        agency_name = ''
+        license_plate = ''
+        duration = 0
+        description = ''
+        location = ''
+        picture_url = ''
+
+        step = handle_start_report(step, twiml)
+
+    elif step == 1:
```
rileynwong: Would suggest against use of magic numbers here (`1, 2, 3, etc.`, see [stack overflow page](http://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad)) - maybe define some constants that correspond to each step. e.g. `STEP_INIT`, `STEP_AGENCY_NAME`, `STEP_VEHICLE_ID`, etc. (or whatever works best for the respective data you're storing)

app/main/messaging.py
```diff
@@ -0,0 +1,392 @@
...
+    set_cookie(response, 'description', description)
+    set_cookie(response, 'location', location)
+    set_cookie(response, 'picture_url', picture_url)
+
+    return response
+
+
+def handle_create_report(agency_name, description, duration, license_plate,
+                         location, picture_url, vehicle_id, phone_number):
+    lat, lon = geocode(location)
```
rileynwong, line 135: nit: i thing `lng` or `long` may be slightly easier to understand than `lon`

app/main/messaging.py
```diff
@@ -0,0 +1,392 @@
...
+
+def handle_location_step(body, step, twiml):
+    validator_form = IncidentReportForm()
+    errors = data_errors(form=validator_form, field=validator_form.location,
+                         data=body)
+
+    if geocode(body) == (None, None):
+        errors.append('We could not find that location. Please respond with a '
+                      'full address including city and state.')
+    if len(errors) == 0:
```
rileynwong, line 181: nit: would `if not errors:` work the same here? if so, i might use that instead bc less redundant and slightly cleaner. but you should test to make sure cuz otherwise you can leave it

app/main/messaging.py
```diff
@@ -0,0 +1,392 @@
...
+
+    else:
+        agency_name = ''
+        twiml.message('Please enter a letter A through {}.'
+                      .format(letters[-1]))
+
+    return agency_name, step
+
+
+def handle_other_agency_step(body, step, twiml):
```
rileynwong, line 226: suggestion: i might leave out the `_step` off the function variable names, e.g. `handle_agency(...)`

edit: actually, can you add a comment or docstring? "handle_other" is a bit confusing, i definitely thought it meant "other agency step" (like part 2 of the agency step) rather than "Other agency" step

app/assets/scripts/map.js
kyle-rosenbluth: I think handling date logic manually can be a bit risky. Why not use `EndDate = Date + timedelta(days=1)` instead? Much less code, and takes care of all of the intricacies and oddities of dates automatically.

app/main/messaging.py
```diff
@@ -0,0 +1,392 @@
...
+    )
+    db.session.add(new_incident)
+    db.session.commit()
+
+    return new_incident
+
+
+def handle_start_report(step, twiml):
+    step = 1
+    twiml.message('What is your location? Be specific! (e.g. "34th and '
```
rileynwong, line 167: suggestion: may consider setting strings as variables, e.g. `PROMPT_MSG_LOCATION`

app/assets/scripts/map.js
rileynwong: Wait, what exactly are we supposed to set the end date to? The current day? Or "tomorrow"? - now that I'm reading your comment, I'm not actually sure what this function is used for lol. What does `getNextDate` mean?/what is "next"?

I'd maybe consider renaming the function and definitely would add a docstring or comment here!

kyle-rosenbluth: I agree, I'm not quite sure what it's supposed to do.

rileynwong:
also lol these temp commits are killing me. if you feel inclined / comfortable enough with rebasing or cherry-picking, i'd make a local copy of this branch and rebase it to clean up the commit history :) rebase + squashing commits would probably be easier than cherry-picking, but that could work also.

app/main/messaging.py
rileynwong, line 181: (nit cont.) but if DOES work the same / has been tested, i would find + replace instances of `if len(errors) == 0` with `if not errors`

app/assets/scripts/map.js
rileynwong: Hmm looks like @sanjayss34 wrote this?

sandlerben: Yeah, I think he accidentally pushed this code here. Please review this code at #83

app/assets/scripts/map.js
sandlerben: Please review this code at #83

sandlerben:
@nancywong I made a noble attempt to clean up the git history and failed. It turns out that when you rewrite the git history with `git rebase -i` you have to re-resolve EVERY merge conflict that occurred (which is insane). I actually attempted to do this, but things got pretty screwed up so I `reset` back.

[Interesting article](https://medium.com/@porteneuve/fix-conflicts-only-once-with-git-rerere-7d116b2cec67#.3nqt3ubhm) on how to use the `git rerere` feature so git remembers how you solve merge conflicts. This can help us a lot in the future.

app/main/messaging.py
sandlerben, line 135: I disagree actually, plus I think with us `lon` elsewhere

app/main/messaging.py
sandlerben, line 181: It would probably work the same but I think it's much better to be explicit here. 

See the debate [here](http://stackoverflow.com/a/7302987/3624201)

app/main/messaging.py
sandlerben, line 226: Yeah good call on docstrings

app/main/messaging.py
sandlerben: Good call

sandlerben:
@nancywong look good now?

sandlerben:
This might be just what we're looking for http://stackoverflow.com/questions/15915430/what-exactly-does-gits-rebase-preserve-merges-do-and-why

Edit: nope, that won't work. See my comment above about rerere.

PR #32, opened 2015-11-11
Admin edit static pages

sandlerben:
r @nancywong @kyle-rosenbluth @aharelick


app/models/__init__.py
```diff
@@ -6,3 +6,4 @@
 from agency import *  # noqa
 from user import *  # noqa
 from incident_report import *  # noqa
+from miscellaneous import *  # noqa
```
rileynwong, line 9: why `miscellaneous` and not `editableHTML`? what other misc classes would potentially be here in `miscellaneous.py`?

app/admin/views.py
```diff
@@ -231,3 +231,24 @@ def change_agency_public_status(agency_id):
     form.process()
     return render_template('admin/manage_agency.html', agency=agency,
                            form=form)
+
+
+@admin.route('/_update_editor_contents', methods=['POST'])
```
aharelick, line 236: I'm assuming this is proper practice to start the route with an underscore?

app/admin/views.py
```diff
@@ -231,3 +231,24 @@ def change_agency_public_status(agency_id):
...
+@login_required
+@admin_required
+def update_editor_contents():
+    """Update the contents of an editor."""
+
+    edit_data = request.form.get('edit_data')
+    editor_name = request.form.get('editor_name')
+
+    editor_contents = EditableHTML.query.filter_by(
```
aharelick: why aren't you using the `get_editable_html()` helper method for this query?

app/templates/macros/page_macros.html
```diff
@@ -0,0 +1,49 @@
+{% macro render_inline_editor(editable_html_obj, current_user) %}
+    <div id="editor-{{ editable_html_obj.editor_name }}" contenteditable="false">
+        {{ editable_html_obj.value|safe }}
+    </div>
+
+    {% if current_user.is_admin() %}
+       <button class="ui button start-edit">
+        Edit
```
aharelick: Nit: should this be indented? you could probably put the whole thing on one line...
`<button class="ui button start-edit">Edit</button>`

app/templates/macros/page_macros.html
```diff
@@ -0,0 +1,49 @@
...
+       </button>
+       <button class="ui primary button end-edit">
+        Save
```
aharelick: Same comment as above about indentation?

app/templates/macros/page_macros.html
```diff
@@ -0,0 +1,49 @@
...
+       </button>
+    {% endif %}
+
+    <script>
+    var editorIDName = "editor-{{ editable_html_obj.editor_name }}";
```
aharelick: might want to indent everything inside the script tag by one 'tab'

app/templates/macros/page_macros.html
```diff
@@ -0,0 +1,49 @@
...
+        });
+
+        $(".end-edit").click(function() {
+            if ( CKEDITOR.instances[editorIDName] ) {
+                var json_data = {
+                    csrf_token: "{{ csrf_token() }}",
+                    editor_name: "{{ editable_html_obj.editor_name }}",
+                    edit_data: CKEDITOR.instances[editorIDName].getData(),
+                };
+                $.post("{{ url_for('admin.update_editor_contents') }}", json_data);
```
aharelick: you might want to post some kind of flash if the update was successful and display something else if anything other than 200 was returned?

aharelick:
@sandlerben, this looks really awesome and clean. I just had a couple of formatting comments and then one question about giving the user feedback after the POST request responds. Besides that looks pretty cool.

app/admin/views.py
sandlerben, line 236: I learned the style from @weronica but I think it's a good practice so it's obvious which routes are not meant to be accessed by the user

aharelick, line 236: :+1:

app/models/__init__.py
sandlerben, line 9: @nancywong Not sure but seemed a bit odd to give this one model it's own python file. What do you think?

sandlerben:
@nancywong @kyle-rosenbluth @aharelick did another round of edits, let me know if it looks good now

app/models/__init__.py
rileynwong, line 9: no strong feelings so otherwise lgtm

rileynwong:
:ship:

PR #38, opened 2015-11-20
Created many-to-many relationship between users and agencies

sandlerben:


sandlerben:
r @nancywong @sanjayss34 @kyle-rosenbluth

manage.py
```diff
@@ -89,7 +89,7 @@ def setup_dev():
                   .filter_by(permissions=Permission.AGENCY_WORKER)
                   .first(),
                   confirmed=True)
-    worker.agency = Agency.query.filter_by(name='SEPTA').first()
+    worker.agencies = [Agency.query.filter_by(name='SEPTA').first()]
```
kyle-rosenbluth, line 92: Could you explain what's going on here? My instinct would be that we should have a constant defined for the `SEPTA` string, but I'm not quite sure what purpose it's serving so I could be wrong.

tests/test_user_model.py
```diff
@@ -141,13 +141,14 @@ def test_anonymous(self):
     def test_agency_worker_role(self):
         Role.insert_roles()
         r = Role.query.filter_by(permissions=Permission.AGENCY_WORKER).first()
-        a = Agency(name='SEPTA')
+        ag_1 = Agency(name='SEPTA')
+        ag_2 = Agency(name='STREETS')
```
kyle-rosenbluth, line 145: Same thought with constants for these.

manage.py
sandlerben, line 92: I think this is okay. We are always going to have a SEPTA object in the db when we get here, and we only have this for debugging locally. This is the default agency worker^

tests/test_user_model.py
sandlerben, line 145: I think this is also fine. We are creating an agency with an arbitrary name. It doesn't really matter what name we give it (just like it's fine for us to create a test user named 'Joe').

sanjayss34:
Looks good to me

app/models/agency.py
```diff
@@ -15,7 +23,8 @@ class Agency(db.Model):
     # public user sees a report on the map, the report's agency will only be
     # shown if is_public is True.
     is_public = db.Column(db.Boolean, default=False)
-    users = db.relationship('User', backref='agency', lazy='select')
+    users = db.relationship('User', secondary=agency_user_table,
```
rileynwong, line 29: could you add a comment that explains that a user can have multiple agencies and that that's why we have this relation? like instead of just 'Configure many to many relationship' explain what those many's are

rileynwong:
otherwise lgtm

PR #41, opened 2015-11-20
Users related to agencies can see a list of those agencies

sandlerben:


app/templates/account/manage.html
```diff
@@ -26,6 +26,10 @@
         <tr><td>Email address</td><td>{{ user.email }}</td></tr>
         <tr><td>Phone number</td><td>{{ user.phone_number }}</td></tr>
         <tr><td>Account type</td><td>{{ user.role.name }}</td></tr>
+        {% if user.agencies|length > 0 %}
+            <tr><td>Affiliated agencies</td><td>{{ user.agencies|join(', ', attribute='name') }}</td></tr>
```
nancywong: What does this mean?

PR #42, opened 2015-11-20
All users now have fake phone numbers

sandlerben:


rileynwong:
this looks good to me

PR #43, opened 2015-11-21
Mobile Nav

kongsally:


kongsally:
r @thomaslee1911 @nancywong

app/assets/styles/app.scss
```diff
@@ -1,6 +1,6 @@
 // Semantic UI breakpoints
 $mobile-breakpoint:        768px;
-$tablet-breakpoint:        992px;
+$tablet-breakpoint:        956px;
```
sandlerben: Why did you change this?

kongsally: That was the point where the last option (log in) was breaking into two lines because there is not enough space. But I think we can revert it back since we merged account into the dashboard

app/templates/macros/nav_macros.html
```diff
@@ -51,7 +70,7 @@
 {# If `count` and `endpoints` are specified, the endpoints will be put into a
  # secondary menu. `count` should be the string (e.g. 'four') number of endpoints. #}
 {% macro desktop_nav(current_user, endpoints=None, count=None) %}
-    <div class="computer tablet only row">
+    <div class="computer only row">
```
sandlerben: What do tablets see then?

sandlerben: Oh jk I see that you moved tablet up

sandlerben: Why did you do that btw?

sandlerben:
Sally looks good except for a few outstanding questions I commented

kongsally:
The questions you had were regarding the nav for tablet and changing the breakpoints, but those are updated now since we have enough space for the tablet and desktop to use the same nav menu.

PR #44, opened 2015-11-21
Admins can now see and filter by workers' agencies

sandlerben:


app/admin/views.py
```diff
@@ -59,8 +59,9 @@ def registered_users():
     """View all registered users."""
     users = User.query.all()
     roles = Role.query.all()
+    agencies = Agency.query.all()
     return render_template('admin/registered_users.html', users=users,
```
rileynwong, line 63: super small nit but could you add a newline above the `return` for readability?

app/templates/admin/registered_users.html
```diff
@@ -27,6 +27,16 @@ <h2 class="ui header">
                         {% endfor %}
                     </div>
                 </div>
+                <div id="select-agency-wrapper">
+                    <select id="select-agency" multiple="" class="ui dropdown">
+                        <option value="">Select agency affiliations</option>
```
rileynwong, line 32: hm `Select agencies` sounds more natural/intuitive imo. If you really wanna keep the 'affiliations', I'd change to `Select affiliated agencies`. in general I would want it to be consistent. e.g. "Affiliated agencies" _everywhere_ instead of mixing that with "Agency affiliations".

I also might simplify `Affiliated agencies` to just `Agencies` since affiliations are already implied

app/admin/views.py
sandlerben, line 63: Thank you, I like nit-picky comments

app/templates/admin/registered_users.html
sandlerben, line 32: done

PR #46, opened 2015-11-21
Made agency name completely case insensitive

sandlerben:


sandlerben:
r @nancywong @thomaslee1911

thomaslee1911:
:shipit: :ship:

app/models/agency.py
```diff
@@ -31,6 +31,15 @@ class Agency(db.Model):
     incident_reports = db.relationship('IncidentReport', backref='agency',
                                        lazy='joined')
 
+    def __init__(self, **kwargs):
+        super(Agency, self).__init__(**kwargs)
```
rileynwong, line 35: hmm what's Agency's super? how come we do this and not `super(**kwargs)`? if this is standard for python pls ignore, but otherwise add a comment?

rileynwong:
otherwise lgtm

app/models/agency.py
sandlerben, line 35: This is pretty standard for python. The super is the `Model` class.

PR #52, opened 2015-11-25
When user changes account type from worker to something else, they lo…

sandlerben:
…se agency affiliations

### DIFF
--- a/app/admin/views.py
+++ b/app/admin/views.py
@@ -133,8 +133,8 @@ def change_account_type(user_id):
     if form.validate_on_submit():
         user.role = form.role.data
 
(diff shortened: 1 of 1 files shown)


app/admin/views.py
```diff
@@ -132,6 +132,12 @@ def change_account_type(user_id):
     form = ChangeAccountTypeForm()
     if form.validate_on_submit():
         user.role = form.role.data
+
+        # If we change the user from a worker to something else, it should lose
```
rileynwong: nit: can we change `it` to `the user`?

sandlerben:
@nancywong look good now?

PR #54, opened 2015-11-30
Incident reports show_agency_publicly attribute

sandlerben:


sandlerben:
r @kongsally @nancywong

app/admin/forms.py
```diff
@@ -81,10 +81,18 @@ class ChangeAgencyOfficialStatusForm(Form):
 class ChangeAgencyPublicStatusForm(Form):
     is_public = SelectField(
         'Publicly visible',
-        description='The public does not see the agency connected to an '
-                    'idling incident if that agency is not publicly visible. '
-                    'Agencies which are not publicly visible still have '
-                    'anonymized reports which the public can see.',
+        description='If an agency is set as public, then all new incident '
```
rileynwong, line 84: Are we splitting the string bc flake8? I think for something like this, we can have it all one string w/ a `# noqa`, cuz otherwise it would be a big pain to edit the description

rileynwong:
otherwise looks fine to me

kongsally:
looks good to me

app/admin/forms.py
sandlerben, line 84: I think it's actually more readable like this.

PR #57, opened 2015-12-05
View list of incident reports for all and for user

rileynwong:


rileynwong:
r @sandlerben @kyle-rosenbluth

app/models/incident_report.py
```diff
@@ -11,6 +11,10 @@ class Location(db.Model):
     incident_report_id = db.Column(db.Integer,
                                    db.ForeignKey('incident_reports.id'))
 
+    def __repr__(self):
+        # TODO: Show address instead?
```
sandlerben, line 15: Why not show everything?

Where is this being used?

app/models/incident_report.py
```diff
@@ -61,6 +66,7 @@ def flip_coin():
                 date=fake.date_time_between(start_date="-1y", end_date="now"),
                 duration=timedelta(minutes=randint(1, 30)),
                 agency=choice(agencies),
+                user=choice(users),
```
sandlerben, line 69: Hmm, I could have sworn I did this already. Is this branch up to date with master?

sandlerben, line 69: I synced it up with master, but still unsure why this shows as an "addition"

app/models/incident_report.py
rileynwong, line 15: Location used to be one of the columns for viewing all reports, but I took it out. It's currently not used anywhere, but I think it's a helpful function to have so I left it in.

What do you mean by showing everything? What else can we show for location besides lat and long?

app/reports/views.py
```diff
@@ -0,0 +1,40 @@
...
+
+@reports.route('/my-reports')
+@login_required
+def view_my_reports():
+    """View all idling incident reports for this user."""
+    reports = current_user.incident_reports
+    agencies = None
+
+    return render_template('reports/reports.html', reports=reports,
+                           agencies=agencies, is_individual=True)
```
sandlerben: Why not just put `agencies=None` here

sandlerben: What do you mean by `is_individual`?

app/models/incident_report.py
rileynwong, line 69: Umm I am pretty sure I added this. This is not on master: https://github.com/hack4impact/clean-air-council/blob/master/app/models/incident_report.py

sandlerben:
The code looks good but a few minor comments
1. The dashboard dropdown shows "All Reports" and "My Reports" but they do not link to the right page. The actual dashboard page doesn't show either of these options.
2. I'm find with the routes being all "all" and "my-reports" but when let's improve the naming in the dashboard. How about "All reports" and "Reports I submitted"?
3. The my-reports page should only show the search bar, not "Admin's Incident Reports" with the long bar.

app/reports/views.py
rileynwong: hmm `is_mine` sounded strange. Basically it's if the reports are for the user or not. Since we are using the same template for `/all` and for `/my-reports`, I check `is_individual` to handle template logic. Do you have any naming suggestions? I couldn't think of a better name...

Maybe `belongs_to_user` ?

rileynwong:
Mm the dashboard stuff is not included in this PR. Actually I think @kongsally will be linking up the dashboard to the endpoints since she is doing the dashboard stuff?

As for 3. I can include it in the next PR, I don't think putting off merging this to change minor formatting is necessary

kongsally:
Yup! I can do it once this branch is merged

sandlerben:
@kongsally  why don't you just do it on this branch?

app/models/incident_report.py
sandlerben, line 15: `original_user_text`?

app/models/incident_report.py
sandlerben, line 69: I've gone ~crazy~ then

app/reports/views.py
sandlerben: How about `render_without_agency_dropdown=True`?

rileynwong: and `agencies=None` in case we decide to change fxnality to add a dropdown for agencies, easier to change that line to set agencies to something else

kongsally:
I guess I could, but I also had another branch for the dashboard layout and dropdown and was planning to make a PR with the proper links to the reports pages after this one was merged.

rileynwong:
yeah ben i think we should merge this branch so we can "lock" this feature for other people to work off of. if sally branches off of this one before merging, any additional changes i make will make things more frustrating and complicated for other dependencies. 

also I renamed the `is_individual` variable to `show_dropdown` and flipped the bools for clarity!

app/models/incident_report.py
sandlerben, line 15: bump

sandlerben:
1. Sure, you can merge it into master and then Sally can work off of it. The best practice is to never merge anything incomplete, half-baked, or not working into master. Sally could easily branch off this branch if she needs the logic here to do her work.
2. Bump (in case you didn't see): The my-reports page should only show the search bar, not "Admin's Incident Reports" with the long bar.
3. I think there is a bug with the all reports page now. When I first load it, it looks like [this](https://www.dropbox.com/s/rcwo88ma3nrlgi5/Screenshot%202015-12-05%2017.59.11.png?dl=0) and after I hover over the agency filter box the text looks small (see [here](https://www.dropbox.com/s/0vdhjle9pr3t0iq/Screenshot%202015-12-05%2017.59.35.png?dl=0))

app/models/incident_report.py
rileynwong, line 15: What's the benefit of including `original_user_text`? I think this is fine and not worth putting off the merge. If someone wants a more useful repr, they can add it in their own PR. This currently isn't used anywhere anymore, I'm just leaving it in for future convenience so others can easily modify it for their own uses.

sandlerben, line 15: Ok sounds good.

rileynwong:
1. I can include it in the next PR, I don't think putting off merging this to change minor formatting is necessary since it does not affect functionality at all. This was a working build and adding on additional features would just add additional delay.
2. sigh I think something sally wrote for styling is causing this and is not a result of this branch. I think we could have avoided this if we had merged this in sooner before sally worked on styling, hence why i kept pushing to get this merged earlier. we'll look into it but tbh i'm a little frustrated with this getting delayed because of the very kind of conflict i was trying to avoid by wanting to merge this in sooner.

sandlerben:
My job as TL is to teach you all industry best practices. It's true that much doesn't _really_ matter on a pre-production project like this, but I still want to be clear about what would be the best thing to do.

What do you mean by "merged it in earlier"? Like merged it in a few hours ago? I am pretty sure it wasn't something added to master today that caused it since this branch was pretty behind master.

kyle-rosenbluth:
:+1:  Looks good.

PR #59, opened 2015-12-12
Edit individual report information

rileynwong:


app/reports/forms.py
```diff
@@ -45,11 +50,12 @@ class NewIncidentForm(Form):
     date = DateField('Date', default=datetime.date.today(),
                      validators=[InputRequired()])
 
-    duration = IntegerField('Idling Duration (in minutes)', [
-        InputRequired('Idling duration (in minutes) is required.'),
+    # TODO - add support for h:m:s format
```
sandlerben: @nancywong do we want to do this in this PR? Should I add another task on trello?

sandlerben: Looks like that the form is rendered with duration in the `h:m:s` format so it does need to be in this PR.

sandlerben:
Should only admins have access to the /info and /edit_info routes? If that's true, then we should change when we link to those routes on the reports.html page depending on the user logged in.

app/reports/views.py
```diff
@@ -13,28 +18,103 @@ def view_reports():
...
+        flash('Report information updated.', 'form-success')
+    elif form.errors.items():
+        flash_errors(form)
+
+    # pre-populate form
+    form.vehicle_id.default = report.vehicle_id
+    form.license_plate.default = report.license_plate
+    # TODO report.bus_number and LED screen number
+    form.location.default = report.location
+    # TODO change location repr to address
```
sandlerben: I really appreciate your use of TODOs in the code

rileynwong: wait is that sarcastic or are you being nice??

sandlerben: No 100%! It's a great habit

sandlerben:
Note: I'm changing the main Incident_Report_Form to have both a picture_file field and picture_url field. I'm doing this because I'm using the validators in this form in csv parsing and twilio stuff. Someone (probably me) needs to reincorporate master and explicitly write which fields should render in templates.

app/models/incident_report.py
```diff
@@ -13,21 +13,22 @@ class Location(db.Model):
...
     vehicle_id = db.Column(db.String(50))
-    license_plate = db.Column(db.String(16))
+    license_plate = db.Column(db.String(16))  # optional
+    bus_number = db.Column(db.Integer)  # optional
+    led_screen_number = db.Column(db.Integer)  # optional
     location = db.relationship('Location',
                                uselist=False,
                                lazy='joined',
                                backref='incident_report')
     date = db.Column(db.DateTime)  # hour the incident occurred
```
sandlerben: this comment isn't quite right - data is a full datetime.datetime object with precision down to the millisecond

app/reports/forms.py
```diff
@@ -47,16 +47,13 @@ class IncidentReportForm(Form):
     longitude = HiddenField('Longitude')
     location = StringField('Address')
 
-    date = DateField('Date', default=datetime.date.today(),
+    date = DateField('Date (year-month-day)', default=datetime.date.today(),
```
sandlerben: date ought to be a datetime.datetime not a datetime.date

sandlerben: Since someone will want to mark that an incident happened at 3:04 PM on March, 3 2015

sandlerben: Check out http://wtforms.simplecodes.com/docs/0.6.1/fields.html#wtforms.fields.DateTimeField

app/reports/views.py
```diff
@@ -13,28 +23,110 @@ def view_reports():
...
+        report.location.original_user_text = form.location.data
+
+        report.date = form.date.data
+
+        # reformat duration data - minutes to time-delta
+        report.duration = minutes_to_timedelta(form.duration.data)
+
+        report.agency = form.agency.data
+
+        # TODO upload picture_file
```
sandlerben, line 105: Will we ever actually have to upload a picture when _editing_ a report?

app/reports/views.py
```diff
@@ -13,28 +23,110 @@ def view_reports():
...
+    """Delete a report"""
+
+    report = IncidentReport.query.filter_by(id=report_id).first()
+    db.session.delete(report)
+    db.session.commit()
+    flash('Successfully delete report.')
 
-    return render_template('reports/reports.html', reports=reports,
-                           agencies=agencies, show_dropdown=False)
+    # TODO flash after redirect
```
sandlerben: Can we remove this TODO?

app/utils.py
```diff
@@ -32,6 +34,25 @@ def parse_phone_number(phone_number):
     return stripped
 
 
+def timedelta_to_minutes(duration):
+    """Convert timedelta object to minutes"""
+    minutes = duration.seconds / 60
```
sandlerben: Do we want floored division always? If so, I would change `/` to `//`

app/utils.py
```diff
@@ -32,6 +34,25 @@ def parse_phone_number(phone_number):
...
+
+    return minutes
+
+
+def minutes_to_timedelta(minutes):
+    return timedelta(minutes=minutes)
+
+
+def flash_errors(form):
```
sandlerben, line 52: Is `utils` really the best place for this?

sandlerben:
@nancywong can you explain what you did with form flashing?

sandlerben:
When I load the info page, I see empty strings in the `License Plate` row and `None` in the `Bus Number` and `LED Screen Number` rows

rileynwong:
That is the expected functionality. Some reports have empty License Plate fields (is it non-optional? i can change the fake report info to always have a license plate). Bus Number and LED Screen Number are both optional fields.

sandlerben:
All three are non optional. I don't understand why for some optional fields it shows the empty string and for others it shows `None`. I would show the empty string for all of them.

rileynwong:
It's because in our fake data, we set it to the empty string for License Plate but not for the other two. Bus Number and LED Screen Number have to be optional, not all vehicles have it?

sandlerben:
On both the /reports/all page and the /info page
- Format date to [this format](https://github.com/hack4impact/clean-air-council/pull/71/files#diff-560eaaac759fd3f33a01ae0b21a509ecR6)
- Change `Duration` headings to `Duration (h:m:s)`
- Rename `Date` headings to `Time` (if you don't think that makes more sense, let me know. That's how I did it here https://github.com/hack4impact/clean-air-council/pull/71/files#diff-560eaaac759fd3f33a01ae0b21a509ecR6)

app/utils.py
rileynwong, line 52: where else would you suggest putting these? and `flash_errors` will show all of the errors at the top of the form at once, rather than showing them in succession. i'll add a docstring

sandlerben:
Gotcha. Can you add some jinja logic so that if the field is empty or None it shows the empty string? I think `None` looks kinda ugly (though this isn't a huge deal).

app/models/incident_report.py
rileynwong: `date`? i'll change the comment

sandlerben:
From my understanding, Alex's submit report code only shows the `Bus Number` and `LED Screen Number` fields if the agency is SEPTA Bus (I think). Can you copy that code to the edit report form?

app/models/incident_report.py
rileynwong: actually, i'll just delete it. i don't think it adds any additional info

app/templates/reports/manage_report.html
```diff
@@ -0,0 +1,101 @@
...
+                                <label>I understand that this action cannot be undone.</label>
+                            </div>
+                        </div>
+                        <a class="ui disabled negative deletion button">
+                            Delete this report
+                        </a>
+                    </div>
+                    </div>
+                {% elif form %}
+                    {{ f.render_form(form) }}
```
sandlerben, line 100: I think this is where you render the `EditIncidentReportForm`. We need to render each field individually because we do not want to show the `picture_file` field here. Unless you think we should support both options in this form?

sandlerben:
It doesn't seem like anything is flashed after a report is deleted

Nancy: lol why can't i just reply to this comment? anyway it should be supported now, i think i may have pushed that commit right after you pulled
image

sandlerben:
Also commented this a few days ago but I'll repost now that I'm doing a real review

Should only admins have access to the /info and /edit_info routes? If that's true, then we should change when we link to those routes on the reports.html page depending on the user logged in.

app/templates/reports/manage_report.html
rileynwong, line 100: hmm i think supporting file upload would be pretty doable. we would have to have the functionality either way, right? i feel like it would actually be easier to support it bc we would already have it for creating a new report anyway

sandlerben, line 100: good point, let's leave both for now

sandlerben:
Can you add a # TODO in the code to automatically hide/show the SEPTA Bus only fields when we add the logic to the create incident form?

app/templates/reports/manage_report.html
```diff
@@ -0,0 +1,106 @@
...
+        <tr><td>Picture</td><td><img src="{{ report.picture_url }}" alt="report picture" width="200"</td></tr>
+        <!-- TODO add width attr to styles -->
+        <tr><td>Description</td><td>{{ report.description }}</td></tr>
+    </table>
+{% endmacro %}
+
+{% block content %}
+    <div class="ui stackable centered grid container">
+        <div class="twelve wide column">
+            <a class="ui basic compact button" href="{{ url_for('reports.view_reports') }}"> <!-- TODO my reports -->
```
sandlerben: Make sure to go through all remaining TODOs you added and decide which need to be in this pr

app/templates/reports/manage_report.html
```diff
@@ -0,0 +1,106 @@
...
+                <i class="caret left icon"></i>
+                Back to all reports
+            </a>
+            <h2 class="ui header">
+                {{ '%s\'s report on %s' % (report.user.full_name(), report.agency.name) }}
```
sandlerben, line 66: Note: not everyone report will have a user! For example, if it is added anonymously via twilio or from the csv.

sandlerben:
Made a few more comments. After those and after we figure out who gets to click on a row to go to the /info page, it is good to merge.

app/reports/views.py
```diff
@@ -13,28 +24,146 @@ def view_reports():
...
+    report = IncidentReport.query.filter_by(id=report_id).first()
+    report_user_id = report.user_id
+
+    db.session.delete(report)
+    db.session.commit()
+    flash('Successfully deleted report.', 'success')
 
-    return render_template('reports/reports.html', reports=reports,
-                           agencies=agencies, show_dropdown=False)
+    # TODO - address edge case where an admin clicks on their own report from
```
sandlerben, line 164: @nancywong how can we handle this case?

requirements/common.txt
```diff
@@ -26,6 +26,8 @@ itsdangerous==0.24
 webassets==0.10.1
 wsgiref==0.1.2
 requests==2.8.1
+WTForms-Components==0.9.9
+pytimeparse==1.1.5
 redis==2.10.5
 rq==0.5.6
--e git://github.com/ui/rq-scheduler.git@4e6837bb6a6e5be4ca364ff9d6218c2b2153009a#egg=rq-scheduler
\ No newline at end of file
+-e git://github.com/ui/rq-scheduler.git@4e6837bb6a6e5be4ca364ff9d6218c2b2153009a#egg=rq-scheduler
```
sandlerben, line 33: Do you know why git thinks this is a change?

sandlerben:
Left one more small comment, after that lgtm

app/reports/views.py
rileynwong, line 164: re: admin redirect to `/all` and not `/my-reports` - i actually don’t know, i left the TODO in there cuz i’m not sure. one of the reasons it took me an extra day to do this was because my first approach involved passing in a boolean variable, `user_only_reports`, that would lmk whether or not we were looking at a report that belonged to the current user. but i couldn’t find a way to do this without passing it thru the URL somehow (e.g. `/reports/13/True/info`), which a pretty glaring security risk. I realized a better way to do it would be to check the report’s user id instead, so I changed it, but am left with the edge case.

Do you know if there’s a way to pass a variable between routes without it being tied to the URL? 
For context, in `manage_report.html`, we link to a specific report from the `/all` or `/my-reports` route using this line:
`{% set href = url_for(route, report_id=report.id) %}` - I could pass in a variable here to signal which route we came from, but I’m not sure how to do that without it being tied to the url (which can easily be modified and bypassed).
There probably is a way but I do not know it. LMK if you can think of anything!

requirements/common.txt
rileynwong, line 33: not sure... newline weirdness possibly? :/ is that a thing?

app/reports/views.py
sandlerben, line 164: We could do something like that with a cookie. 

Leave this TODO, but mention that you think it can be solved with a cookie.

PR #62, opened 2015-12-12
Add geocoding function to utils.py

thomaslee1911:
This refactors out the geocoding logic from parse_csv, including viewport-biasing base on the default viewport parameter. @kyle-rosenbluth @sanjayss34 @sandlerben


app/utils.py
```diff
@@ -29,3 +31,16 @@ def parse_phone_number(phone_number):
         stripped = '1' + stripped
     stripped = '+' + stripped
     return stripped
+
+
+# Viewport-biased geocoding using Google API
+# Returns a tuple of (latitude, longitude), (None, None) if geocoding fails
+def geocode(address):
+    url = "https://maps.googleapis.com/maps/api/geocode/json"
+    payload = {'address': address, 'bounds': config['default'].VIEWPORT}
```
sandlerben: This isn't how you should get things from config. You should get stuff from app.config

app/utils.py
```diff
@@ -29,3 +31,16 @@ def parse_phone_number(phone_number):
...
+    r = requests.get(url, params=payload)
+    if r.json()['status'] is 'ZERO_RESULTS' or len(r.json()['results']) is 0:
+        return (None, None)
```
sandlerben: Nit: don't include redundant parenthesis around tuples

app/utils.py
```diff
@@ -29,3 +31,16 @@ def parse_phone_number(phone_number):
...
+    else:
+        coords = r.json()['results'][0]['geometry']['location']
+        return (coords['lat'], coords['lng'])
```
sandlerben: same as above

sandlerben:
Had a few small comments, besides that lgtm

app/utils.py
thomaslee1911: Got it

app/utils.py
thomaslee1911: Ok

thomaslee1911:
@kyle-rosenbluth it is merged.

PR #65, opened 2016-01-05
Admins can now add agencies (which are automatically official)

sandlerben:


app/templates/admin/add_agency.html
```diff
@@ -0,0 +1,33 @@
+{% extends 'layouts/base.html' %}
+{% import 'macros/form_macros.html' as f %}
+
+{% block content %}
+    <div class="ui grid container">
+        <div class="eight wide computer sixteen wide mobile centered column">
+            <h2 class="ui dividing header">Create an agency</h2>
+
+            {% set flashes = {
```
rileynwong, line 9: not particularly pressing/something to consider - should we refactor the flashes stuff if we're using repeated code? 
`git grep -i "set flashes"` to see everywhere it's used, which is at least 4 times in our code base

rileynwong:
otherwise lgtm

app/templates/admin/add_agency.html
sandlerben, line 9: I don't mind being explicit here, but if you want I welcome a change like that

PR #66, opened 2016-01-05
Add weather data

sandlerben:


app/templates/reports/manage_report.html
```diff
@@ -30,6 +30,13 @@
         <tr><td>Duration</td><td>{{ report.duration}}</td></tr>
         <tr><td>Agency</td><td>{{ report.agency.name }}</td></tr>
         <tr><td>Picture</td><td><img src="{{ report.picture_url }}" alt="report picture" width="200"</td></tr>
+        <tr><td>Weather During Incident</td>
```
rileynwong: would suggest "Weather conditions during incident" for clarity

app/templates/reports/manage_report.html
```diff
@@ -30,6 +30,13 @@
...
+            <td>
+                {% for line in report.weather.splitlines() %}
+                {{ line }}<br>
```
rileynwong: nit: indent this line?

app/templates/reports/reports.html
```diff
@@ -70,6 +70,7 @@ <h2 class="ui header">
                             <th>Vehicle ID</th>
                             <th>License Plate</th>
                             <th>Duration</th>
+                            <th>Weather During Incident</th>
```
rileynwong: as above, would suggest "Weather conditions during incident" for clarity

app/utils.py
```diff
@@ -51,3 +53,44 @@ def geocode(address):
     else:
         coords = r.json()['results'][0]['geometry']['location']
         return coords['lat'], coords['lng']
+
+
+def get_current_weather(location):
+    """TODO write docstring"""
```
rileynwong: What are the minimum conditions or exceptions for idling incidents? Wondering what conditions are considered relevant data

config.py
```diff
@@ -23,6 +23,9 @@ class Config:
     # Default viewport is bounding box for Philadelphia, PA
     VIEWPORT = '39.861204,-75.310357|40.138932,-74.928582'
 
+    OPEN_WEATHER_MAP_API_KEY = os.environ.get('OPEN_WEATHER_MAP_API_KEY') or \
+        'SjefBOa$1FgGco0SkfPO392qqH9'
```
rileynwong, line 29: is this our api key?? probably shouldn't commit this. once we merge in this branch we should let everyone know to add the api key to our local configs

rileynwong:
otherwise lgtm

config.py
sandlerben, line 29: nah this is a random string

app/utils.py
sandlerben: @nancywong What do you mean by this?

sandlerben: I don't think this has to do with the code so I'm going to merge

rileynwong: oh, as in - are temperature, pressure, and humidity the conditions that are important to the cac? i think temperature is definitely, but unsure about the others. might want to check with them what is and is not important. in any case this is more wrt product design than the code itself.

app/templates/reports/manage_report.html
```diff
@@ -30,6 +30,13 @@
         <tr><td>Duration</td><td>{{ report.duration}}</td></tr>
         <tr><td>Agency</td><td>{{ report.agency.name }}</td></tr>
         <tr><td>Picture</td><td><img src="{{ report.picture_url }}" alt="report picture" width="200"</td></tr>
+        <tr><td>Weather conditions during incident</td>
```
kongsally, line 45: I think we can just call it "Weather Condition" since we know it's already for an incident report

app/templates/reports/reports.html
```diff
@@ -70,6 +70,7 @@ <h2 class="ui header">
                             <th>Vehicle ID</th>
                             <th>License Plate</th>
                             <th>Duration</th>
+                            <th>Weather conditions during incident</th>
```
kongsally, line 75: Same here "Weather Condition"

kongsally:
Small detail, but after that change, lgtm

app/templates/reports/reports.html
sandlerben, line 75: Lol Nancy had the exact opposite comment

@nancywong @kongsally what should I do?

rileynwong, line 75: oh, my comment was to add the word "condition". i don't have very strong feelings about this, but i think being explicit is better. if it looks terrible then i'm ok with shortening it, but otherwise i don't see the harm in being explicit.

PR #68, opened 2016-01-09
Added a rq task queue

sandlerben:


run-worker.py
```diff
@@ -0,0 +1,22 @@
+"""Initializes a slim rq task queue. File taken from
+http://python-rq.org/patterns/"""
+
+import os
+import urlparse
+from redis import Redis
+from rq import Worker, Queue, Connection
+
+listen = ['high', 'default', 'low']
```
jondubin, line 9: Are you using three queues/should we have three in Flask base?

sandlerben, line 9: I think it's standard to use these three queues but flask-base might only need the default queue.

jondubin, line 9: Yeah it seems relatively non-consequential either way. I'm guessing in practice the expectation is that you place tasks in the queue according to their priority. I can't imagine that for the time being we'll have this need.

PR #71, opened 2016-01-10
Refactor async email and worker notify email

sandlerben:


app/email.py
```diff
@@ -1,21 +1,16 @@
...
-def send_async_email(app, msg):
+def send_async_email(recipient, subject, template, **kwargs):
+    app = create_app(os.getenv('FLASK_CONFIG') or 'default')
     with app.app_context():
+        msg = Message(app.config['EMAIL_SUBJECT_PREFIX'] + ' ' + subject,
+                      sender=app.config['EMAIL_SENDER'],
+                      recipients=[recipient])
+        msg.body = render_template(template + '.txt', **kwargs)
+        msg.html = render_template(template + '.html', **kwargs)
         mail.send(msg)
```
jondubin, line 16: Is it appropriate to call this function `send_async_email`? It looks entirely synchronous

sandlerben, line 16: That's a good point. This function is meant to be used off of the main thread (hence why I create a new flask app) but in theory it could be used synchronously.

sandlerben, line 16: You think I should rename it?

jondubin, line 16: When you call this function (from wherever), it executes synchronously, right? Before with `send_email`, actually sending the message was relegated to a separate thread which the function `send_email` only started. But now when you call `send_async_email` it creates the app, ravages it for config vars, and then sends the message, but it's still getting executed in the same thread as whatever called it. 

When I see `async`, I think, if I call this function, execution will resume before the email has actually been sent. That doesn't seem to the case here?

sandlerben, line 16: renamed

rileynwong:
lgtm

PR #72, opened 2016-01-10
Added upload_image function to upload an image to Imgur

sandlerben:


app/utils.py
```diff
@@ -43,3 +44,32 @@ def geocode(address):
...
+        description = 'This is part of an idling vehicle report on {}.'.format(
+            current_app.config['APP_NAME'])
+
+    if image_url is not None:
+        result = client.upload_from_url(url=image_url, config={
+            'title': title,
+            'description': description,
+        })
+    else:
+        result = client.upload_from_path(path=image_file_path, config={
```
rileynwong, line 77: what's the format of `result`? is it a url? if not, we might want to add a helper that will parse the result and return the imgur_url

sandlerben, line 77: It's a dictionary. I agree that we should wrap it in something or return a tuple of the url and anything else needed.

PR #78, opened 2016-01-13
Add rq scheduler

sandlerben:


sandlerben:
@jondubin or @aharelick can you take a look? How do you feel about adding a rq-scheduler to flask-base?

manage.py
```diff
@@ -129,5 +135,47 @@ def setup_general():
...
+        db=0,
+        password=app.config['RQ_DEFAULT_PASSWORD']
+    )
+
+    setup_loghandlers('INFO')
+    scheduler = Scheduler(connection=conn, interval=60.0)
+    for _ in xrange(10):
+        try:
+            scheduler.run()
+        except ValueError, exc:
```
jondubin: Nit: can you use `as` syntax instead of the comma here? Actually hadn't seen the comma syntax before. http://stackoverflow.com/questions/2535760/python-try-except-comma-vs-as-in-except

manage.py
```diff
@@ -129,5 +135,47 @@ def setup_general():
...
+            if exc.message == 'There\'s already an active RQ scheduler':
+                scheduler.log.info(
+                    'An RQ scheduler instance is already running. Retrying in '
+                    '%d seconds.', 10,
+                )
+                time.sleep(10)
+            else:
+                raise
```
jondubin: A naked raise, does this work? Should we raise the original exception `exc`?

PR #80, opened 2016-01-15
Made some small changes to the report info box on the map page

sandlerben:


app/templates/main/map.html
```diff
@@ -57,14 +57,16 @@
...
                                 '</button>' +
-                                '<h3>Date: ' + '{{ report.date }}' + '</h3>' +
-                                '<p>Duration: ' + '{{ report.duration }}' + '</p>' +
+                                '<h3>Date: ' + '{{ report.date.strftime('%m-%d-%Y at %I:%M %p') }}' + '</h3>' +
+                                '<p>Duration (h:m:s): ' + '{{ report.duration }}' + '</p>' +
                                 '<p>Vehicle ID: ' + '{{ report.vehicle_id }}' +'</p>' +
-                                '<p>License Plate: ' + '{{ report.license_plate }}' + '</p>' +
-                                {% if current_user.is_admin() or report.agency in current_user.agencies or report.agency.is_public %}
+                                '<p>License Plate: ' + '{{ report.license_plate if report.license_plate else '' }}' + '</p>' +
+                                {% if current_user.is_admin() or report.agency in current_user.agencies or report.show_agency_publicly %}
```
rileynwong, line 64: Why the distinction between `is_public` and `show_agency_publicly`? Do we have agencies that are public that we shouldn't be showing publicly? If yes, I would add a docstring in `models/incident_reports.py`. If no, then... well either way, I would add a docstring why the distinction exists.

Looking through `models/incident_reports.py`, looks like the difference is if the report is public vs. the agency, right? I might suggest changing `show_agency_publicly` to something like `is_public_report` or just `is_public`.
Or is the report being shown but the agency name being hidden/kept private? In which case, would suggest `report.has_public_agency` or `report.agency_is_public`

overall would suggest a change in variable name + comment string in `models/incident_reports.py`

sandlerben, line 64: > Or is the report being shown but the agency name being hidden/kept private? In which case, would suggest report.has_public_agency or report.agency_is_public

Yes it's this. I'll take a look at the documentation but I personally think the variable name is fine.

PR #83, opened 2016-01-19
Date slider fix

sandlerben:
@sanjayss34 let's review this code here before incorporating it anywhere else

### DIFF
--- a/app/assets/scripts/map.js
+++ b/app/assets/scripts/map.js
@@ -158,37 +158,8 @@ $(document).ready(function() {
 });
 
 function getNextDate(startDate) {
-    date = startDate.getDate();
-    month = startDate.getMonth();
-    year = startDate.getFullYear();
-    MAX_MONTH = 11;
-    monthDays = {
-        0: 31,
-        1: 28,
-        2: 31,
-        3: 30,
-        4: 31,
-        5: 30,
-        6: 31,
-        7: 31,
-        8: 30,
-        9: 31,
-        10: 30,
-        11: 31
-    }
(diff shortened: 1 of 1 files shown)


sandlerben:
See comments from #29 [here](https://www.dropbox.com/sh/pgik4wxz94ztotp/AACYx5Q6jc0fI9IRk117_jjYa?dl=0)

sanjayss34:
Thanks @nancywong and @kyle-rosenbluth for the good comments; I'm sorry for the very delayed response!

app/assets/scripts/map.js
```diff
@@ -157,6 +157,11 @@ $(document).ready(function() {
     });
 });
 
+function getNextDate(startDate) {
```
sandlerben, line 160: Can you explain what this does and why it's needed?

sanjayss34, line 160: This function returns the next date after the given date. The way it's used for setting the date bounds for incident reports is: I pass the end date that the user specifies with the date slider to this function, get back the date after that end date (i.e. the date corresponding to 12:00 AM on the following day) and then put incident reports on the map iff the date of the incident report is before this new date. For instance, if the user specifies January 26, 2015 as the end date, I check the incident reports to see that they were placed before January 27, 2015 at 12:00 AM. The reason to use the next day is that I don't want to filter out any reports that were registered at some time on the end date (in this example, those reports made on January 26). Do you think I should use a different approach?

sandlerben, line 160: Nope that's great.

